PR #13, opened 2021-01-06
Setup eslint and run

zakpatterson:
I suggest we set up eslint so we both and future contributors use a common code style. [I've picked the 'standard' option at this point](https://standardjs.com/#why-should-i-use-javascript-standard-style). It's just a suggestion, let me know if that's okay.


src/components/createPDF.js
```diff
@@ -1,19 +1,19 @@
...
             <form>
-                <Box display="flex" justifyContent="flex-start">   
+                <Box display="flex" justifyContent="flex-start">
                     <h2>Print Copy to File</h2>
                 </Box>
 
                 <Box display="flex" justifyContent="flex-start" paddingTop={2} paddingBottom={1}>
                     <Box display="flex" justifyContent="flex-start" paddingRight={2}>
-                        <Button component={Link} to={"familyinfo"} variant="contained" color="secondary" >
+                        <Button component={Link} to={'familyinfo'} variant="contained" color="secondary" >
```
zakpatterson, line 16: Probably not in `{ }`s.

zakpatterson, line 16: There are a number of these where a plain string is inside jsx `{}`. Is there a reason?

thegrims: That's a good callout. I didn't realize that the link strings could be added without the curly braces but that totally makes sense given that it's static rather than a variable.

zakpatterson: @thegrims I added another commit to remove all those plain strings in {}s

thegrims: Awesome, thanks!

PR #35, opened 2021-01-10
Fix a few issues in display of dropdown

### LINKED ISSUE #17: Change dropdown format to allow for easy navigation while being performant with many elements
> The state dropdown is... confusing. Under the hood it's a bunch of <option>s like I'd expect, but they're popped out into some div, I guess so you can style them? Except this doesn't replicate the functionality of the basic HTML controls -- there's zero keyboard navigation. (To add insult to injury, you record the state as the two-letter thing I want to type anyway, but typing that after I open the dropdown does nothing.) It's not like React can't do a normal dropdown, so what went wrong here?

[https://www.reddit.com/r/opensource/comments/kqr89x/open_source_tax_filing_website/gi6d0jv/](https://www.reddit.com/r/opensource/comments/kqr89x/open_source_tax_filing_website/gi6d0jv/)

zakpatterson:
Uses native property on textfield select to fix rendering on tauri
and improve look on browser.

Also switch order of postal vs full name, for better keyboard nav
as noted in issue.

Also factored out USStateDropdown as separate component to remove
code duplication.

Fixes #17

### DIFF
--- a/src/components/labeledInput.js
+++ b/src/components/labeledInput.js
@@ -2,6 +2,7 @@ import React from 'react'
 import { TextField, Box, Checkbox, FormControlLabel } from '@material-ui/core'
 import { Controller } from 'react-hook-form'
 import InputMask from 'react-input-mask'
+import locationPostalCodes from '../data/locationPostalCodes'
 
 export function LabeledInput ({ strongLabel, label, register, required, mask, pattern, patternDescription, name, errors, defaultValue }) {
   let helperText = ''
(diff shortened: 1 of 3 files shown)


thegrims:
The functionality looks much better! What is the advantage of having USStateDropDown as a separate function rather than calling LabeledDropdown directly in w2EmployerInfo with the props specified in USStateDropDown?

zakpatterson:
A drop down of US states should behave the same everywhere that it is included in the project, and yet in each place it is used, we must repeat the same mappings [like here](https://github.com/thegrims/UsTaxes/pull/35/files#diff-674062f4712ecb72091012124d91438802dcabef0532e26b2824bc58664a3be9L111), and include the same locality file to reference the list of states. Since this code is repeated in two places, it's good to factor that out.

thegrims:
Ok sounds good

PR #48, opened 2021-01-17
Refactor menu to take sections as properties

zakpatterson:
Also refactor the save and continue buttons to take the next url as a parameter. This lets us maintain all the router URLs in one place.

### DIFF
--- a/src/components/menu.tsx
+++ b/src/components/menu.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { ReactElement } from 'react'
 import Divider from '@material-ui/core/Divider'
 import Drawer from '@material-ui/core/Drawer'
 import Hidden from '@material-ui/core/Hidden'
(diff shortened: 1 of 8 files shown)


src/components/w2.js
```diff
@@ -28,27 +28,45 @@ const theme = createMuiTheme({
...
+      ['Employer Information', '/w2employerinfo'],
+      ['Employee Information', '/w2employeeinfo'],
+      ['Job Information', '/w2jobinfo']
+    ]
+  },
+  {
+    title: 'Personal',
+    items: [
+      ['Family Information', '/familyinfo'],
+      ['Review and Print', '/createPDF']
```
thegrims, line 44: /createPDF should be lowercase since it's part of a url

src/components/w2.js
```diff
@@ -28,27 +28,45 @@ const theme = createMuiTheme({
   }
 })
 
+const drawerParts = [
```
thegrims, line 31: maybe could be drawerSections

thegrims (commented):
the family information page is not highlighted when it is navigated to

thegrims:
review and print should be under a separate list with a divider since it's logically separate from family information

src/components/menu.tsx
```diff
@@ -46,91 +46,52 @@ const useStyles = makeStyles((theme) => ({
...
-        >
-          <ListItemText primary='Review and Print' />
-        </ListItem>
-      </List>
+      {
+        sections.map(({ title, items }, i) => (
+          <div key={i}>
+            <h2>{title}</h2>
+            <List>
+              {items.map((x, j) =>
```
thegrims, line 77: should change variable names from x, j to what these variables represent

thegrims:
Might want to add the previous section to props as well, but if you don't want to do that right now, that's totally ok.

zakpatterson:
> Might want to add the previous section to props as well, but if you don't want to do that right now, that's totally ok.

I agree, but I think it's out of scope at the moment. It would also be nice to have the Save, Back, Forward actions as a container component so the Previous, Save, etc buttons would not have to be duplicated on each page.

thegrims:
Looks great! Thanks for your hard work!

zakpatterson:
> Might want to add the previous section to props as well, but if you don't want to do that right now, that's totally ok.

Do you think all the back buttons can just be `history.back()`?

thegrims:
> > Might want to add the previous section to props as well, but if you don't want to do that right now, that's totally ok.
> 
> Do you think all the back buttons can just be `history.back()`?

That's a great idea, yeah that would work

zakpatterson:
> That's a great idea, yeah that would work

Ok, I'll try that, mitigating the issue of just navigating directly to a url.

thegrims:
@zakpatterson After thinking about history.back some more I don't think it would work in the way that would be desired because if someone navigated to the review and print page from Employer Information, the back command would then take them to Employer Information rather than to Family Information

zakpatterson:
You don't think that's what the user would expect? I guess you are thinking about it more in terms of next step, previous step.

thegrims:
Yeah, I was thinking more in the form of next step, previous step. Maybe I'll change back to previous to make it more clear

PR #50, opened 2021-01-18
Add hook to handle page sequencing

zakpatterson:


zakpatterson:
Previous button is still a little broken when the user selects from the menu. Will try to fix soon

src/components/familyInfo.js
```diff
@@ -84,18 +82,7 @@ export default function FamilyInfo ({ nextUrl }) {
...
-            <Button component={Link} to="w2jobinfo" variant="contained" color="secondary" >
-              Back
-            </Button>
-          </Box>
-
-          <Button type="submit" variant="contained" color="primary">
-            Save and Continue
-          </Button>
-        </Box>
+        {children}
```
thegrims, line 85: what does children represent here?

thegrims:
In the deploy preview, it's taking 2 clicks of the browser back button to get to the previous page

src/components/familyInfo.js
zakpatterson: Same as any react component that can have a children property, just children of the `<FamilyInfo>` element. I settled on doing it this way instead of including some parameter (`isLastPage` or something like that), so that all the page usages in `Main.tsx` would specify their own buttons as children. Is it good?

zakpatterson:
> In the deploy preview, it's taking 2 clicks of the browser back button to get to the previous page

Weirdly this wasn't happening locally, and after just merging master back in here to trigger a redeploy, it looks to be fixed? Maybe netlify's deployment was stale somehow? Do you see it working now?

zakpatterson:
Ha ok, really think I got it this time.

src/components/familyInfo.js
zakpatterson: Better in my opinion would be to have a <FormWizard /> element, which steps through the children routes, but I couldn't figure out a clean way to have the <form> element in each page, and the submit button injected into each page's form.

thegrims: > Same as any react component that can have a children property, just children of the `<FamilyInfo>` element. I settled on doing it this way instead of including some parameter (`isLastPage` or something like that), so that all the page usages in `Main.tsx` would specify their own buttons as children. Is it good?

Could you change the variable name from children to navButtons for clarity?

thegrims:
It's working for me as well! Nice

src/components/familyInfo.js
zakpatterson: Sure.

zakpatterson: Done

thegrims:
Looks good! Merging

Issue #18, opened 2021-01-08
Investigate using react-number-format over react-input-mask


thegrims:
https://www.npmjs.com/package/react-number-format

May improve formatting for dollar amounts with commas as well as addressing these issues

> A quick glance at the repo: It looks like you're using JS numbers for currency, a bunch of them explicitly floating-point (in the tax brackets). If you do any math with this, you'll be doing math with floats. This is a bad idea for currency -- see https://0.30000000000000004.com/
> 
> You can probably get away with using integer number of cents with normal JS numbers... I'd be looking into a big-decimal library of some sort instead -- JS seems to have a few, I guess bignumber.js might work?
> 
> I was curious what it'd actually do, so I tried the website... and I know you said "work in progress", so I'm trying to be constructive here... but it was pretty broken:
> 
> Trying to enter an Employer Identification Number, neat how it enters the - and everything, but after filling the page with nonsense data (EIN of '12-34567'), it won't save and continue, and I get "Input should be filled with 7 numbers" on the EIN.
> Oh, it wants 9 numbers. Okay, fair enough, but...
> Now it eats inputs. I type '8' at the end of the line, and that's fine, then I type '9' and it doesn't appear in the 'input' but the box appears to validate... until I tab away.
> Fine, let's clear it and try again... First keystroke gets eaten before I even start typing. So if I type '1234' it gets recorded as '23-4'.
> Basically, had to clear everything, reload the page, and type the entire fake number in one go without any line-editing...

https://www.reddit.com/r/opensource/comments/kqr89x/open_source_tax_filing_website/gi6d0jv/

PR #55, opened 2021-01-20
Add typings for labeled input, and all form pages

zakpatterson:
* Adds types to all form pages
* Adds types to `LabeledInput`
    * Refactors to pass reusable `PatternConfig` objects instead of individual properties
* Updates `react-form-hook` to next version (for types)


thegrims:
I think form fields with no mask should be full width

src/components/TaxPayer.tsx
```diff
@@ -1,21 +1,23 @@
...
 
-export default function TaxPayerInfo ({ navButtons, onAdvance }) {
+export default function TaxPayerInfo ({ navButtons, onAdvance }: PagedFormProps): ReactElement {
   const { register, handleSubmit, errors } = useForm()
   // const variable dispatch to allow use inside function
   const dispatch = useDispatch()
 
-  const prevFormData = useSelector(state => {
-    return state.information.taxpayer ?? {}
+  const prevFormData: TaxPayer | undefined = useSelector((state: TaxesState) => {
```
thegrims, line 15: this can be const prevFormData?: TaxPayer

src/components/W2EmployerInfo.tsx
```diff
@@ -77,7 +77,7 @@ export default function W2EmployerInfo ({ navButtons, onAdvance }) {
           register={register}
           required={true}
           name="employerCity"
-          defaultValue={prevFormData.employerCity}
+          defaultValue={prevFormData?.employerCity ?? ''}
```
thegrims, line 80: Does defaultValue need to be a string? defaultValue could be set as an optionally defined string in the labeledInputProps interface like so 
```
interface LabeledInputProps {
  ...
  defaultValue?: string
}
```

src/components/createPDF.tsx
```diff
@@ -1,11 +1,12 @@
...
 import { createPDFPopup } from '../pdfFiller/fill1040Fields'
+import { DonePagedFormProps } from './pager'
 
-export default function CreatePDF ({ navButtons }) {
-  const onSubmit = (e) => {
+export default function CreatePDF ({ navButtons }: DonePagedFormProps): ReactElement {
+  const onSubmit = (e: FormEvent<any>): void => {
     e.preventDefault()
-    createPDFPopup()
+    createPDFPopup().then(() => {}, () => {})
```
thegrims, line 9: Were these anonymous functions added for a specific reason?

thegrims (changes requested):

package.json
```diff
@@ -20,7 +20,7 @@
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-hook-form": "^6.14.2",
-    "react-input-mask": "^2.0.4",
+    "react-input-mask": "^3.0.0-alpha.2",
```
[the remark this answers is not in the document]
zakpatterson: Don't know for sure, 'alpha'/'next' version has types for ts, 2-series version does not. https://github.com/sanniassin/react-input-mask#known-issues

src/components/createPDF.tsx
zakpatterson: This is used to ignore the effects the promise returned by createPDFPopup(). This is a 'dangling future' I think it's called, since createPDFPopup returns a promise, and we need this function to simply return void. So the TS compiler catches this, but in this case there's not much to be done since the form's `onSubmit` call wants a synchronous function.

src/components/RefundBankAccount.tsx
```diff
@@ -1,21 +1,24 @@
...
+import { PagedFormProps } from './pager'
+
+export default function RefundBankAccount ({ navButtons, onAdvance }: PagedFormProps): ReactElement {
   const { register, handleSubmit, errors } = useForm()
   // const variable dispatch to allow use inside function
   const dispatch = useDispatch()
 
-  const prevFormData = useSelector(state => {
-    return state.information.refund ?? {}
+  const prevFormData: Refund | undefined = useSelector((state: TaxesState) => {
```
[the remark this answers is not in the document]
zakpatterson: I don't think that syntax works? The type of `prevFormData` is `Refund | undefined`. `const x?: string = undefined` is not valid. `const x? = "hi"` is not valid.

src/components/W2EmployerInfo.tsx
zakpatterson: Ah very nice.

src/components/createPDF.tsx
zakpatterson: Added code comment

src/components/TaxPayer.tsx
zakpatterson: `const blah? = __` is not valid. That shorthand is just for interface properties, type declarations, function parameters.

package.json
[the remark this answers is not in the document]
zakpatterson: We already have #18, so we'll probably get rid of this dependency anyway?

src/components/RefundBankAccount.tsx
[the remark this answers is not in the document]
thegrims: You're right haha, good callout. I got carried away with the ? optional parameter syntax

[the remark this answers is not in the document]
zakpatterson: I've tried to write it that way a few dozen times too, the syntax is valid in every other situation I can think of so

package.json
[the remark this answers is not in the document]
thegrims: Yeah good points

src/components/labeledInput.tsx
```diff
@@ -0,0 +1,271 @@
...
+    const r = errors[name]
+    result = r.type
+  }
+  return result
+}
+
+const isError = (errors: Errors, name: string): boolean => getError(errors, name) !== undefined
+
+export function LabeledInput (props: LabeledInputProps): ReactElement {
+  const { strongLabel, label, register, required, patternConfig = {}, name, errors, defaultValue } = props
```
thegrims, line 42: can be replaced with
`export function LabeledInput({ strongLabel = '', patternConfig, label, register, required, name,errors, defaultValue}: LabeledInputProps): ReactElement {`

src/components/labeledInput.tsx
```diff
@@ -0,0 +1,271 @@
...
+        <p><strong>{strongLabel}</strong>{label}</p>
+      </Box>
+      <Box display="flex" justifyContent="flex-start">
+        {input}
+      </Box>
+    </div>
+  )
+}
+
+const id = <A,>(a: A): A => a
```
thegrims, line 94: what's this function doing?

thegrims:
Getting this linter error in W2EmployerInfo onsubmit
image

src/components/labeledInput.tsx
```diff
@@ -0,0 +1,271 @@
...
+import InputMask from 'react-input-mask'
+import locationPostalCodes from '../data/locationPostalCodes'
+
+interface Errors {
+  [key: string]: {type: string}
+}
+
+type Register = (rules?: RegisterOptions) => any
+
+interface PatternConfig {
```
thegrims, line 13: Could these interfaces be factored into a separate file?

src/components/labeledInput.tsx
zakpatterson: Re: linter error, definitely looks like that should be a linter error, trying to figure out why I don't see it. Do you see it if you run `npm run lint`? What version of eslint is your vscode using?

zakpatterson: I reorganized the labeledInput file a little bit so that hopefully it's a little more clear. This is just a default parameter for a LabeledDropdown that only contains an array of strings.

src/components/labeledInput.tsx
zakpatterson: I kind of want to keep it this way? A combination of readability of the type signature and keeping the line from getting too long seems better to me than being concise with destructuring.

src/components/labeledInput.tsx
zakpatterson: :recycle:

zakpatterson:
About the linter error, what version of eslint is your vscode using? Just trying to figure out why I'm not seeing it. Do you see it if you run `npm run lint`?

thegrims:
> About the linter error, what version of eslint is your vscode using? Just trying to figure out why I'm not seeing it. Do you see it if you run `npm run lint`?

Weirdly after reinstalling and running this afternoon it seems like the linter issue disappeared! I think this must have been an issue with my local linter. The eslint version used by vscode should just be the version installed in  my case in package-lock.json right?

src/components/createPDF.tsx
thegrims: On second look this can be changed to
 ```
 const onSubmit = async (e: FormEvent<any>): Promise<void> => {
    e.preventDefault()
    await createPDFPopup()
  }
```
I think this will increase clarity

File: src/components/labeledInput.tsx [hunk 9]
```diff
@@ -0,0 +1,229 @@
...
+        <p><strong>{strongLabel}</strong>{label}</p>
+      </Box>
+      <Box display="flex" justifyContent="flex-start">
+        {input}
+      </Box>
+    </div>
+  )
+}
+
+interface BaseDropdownProps {
```
thegrims, line 89: Can other interfaces like this and LabeledDropdownProps<A,> and LabeledInputProps also be factored into a separate interfaces file

src/components/labeledInput.tsx
thegrims: What does A and a signify? are they just generic types?

zakpatterson: Yeah it explicitly doesn't signify anything, so that's why that naming convention is used.

```
  // this function takes an A which can be anything, and returns an A, so we know it can 
  // only possibly return the input of the function.
  const id = <A,>(a: A): A => a

  // return type tells you this can return any string in the universe
  const doNothing = (optionText: string): string => optionText

  // still hints it could return any string in the universe
  const doNothing2: ((x: string) => string) = id

  // any possible number, who knows
  const doNothing3 = (n: number): number => n
  const doNothing4: ((n: number) => number) = id
```

src/components/labeledInput.tsx
zakpatterson: Yeah this file is doing way too much, I'll break it up.

zakpatterson:
> Weirdly after reinstalling and running this afternoon it seems like the linter issue disappeared! I think this must have been an issue with my local linter. The eslint version used by vscode should just be the version installed in  my case in package-lock.json right?

I'm not sure but I would guess it's the one in your PATH, or in vscode settings.

src/components/RefundBankAccount.tsx
```diff
@@ -1,21 +1,25 @@
-import React from 'react'
+import React, { ReactElement } from 'react'
 import { useForm } from 'react-hook-form'
 import { Box } from '@material-ui/core'
 import { useDispatch, useSelector } from 'react-redux'
-import { LabeledInput } from './labeledInput'
+import { LabeledInput } from './input/LabeledInput'
```
thegrims, line 5: Could this and some of the other input imports be changed to 

`import { LabeledInput } from './input'`

because of the addition of the input index file?

thegrims:
Other than that last comment, looks ready to merge to me!

PR #59, opened 2021-01-23
Taxpayer dependents, spouse, filing status, improved PDF field mapping

zakpatterson:
Add spouses and dependents in Taxpayer form

Refactored actions to be a little simpler to declare.

Also see new way of mapping pdf fields to model data. Each field index maps to a function from the data model to a string.

TODO: 
* :heavy_check_mark:  Need to correct field mappings to PDF
* :heavy_check_mark: Need to add filing statuses (HoH allowed with at least one dependent, MFS/MFJ allowed with spouse)
* :heavy_check_mark:  Fix W2 additions. W2s can come from either spouse, so instead of `W2EmployeeInfo`, we just need a selector for which spouse the W2 comes from. 

This also changes how errors are passed around. Instead of passing the entire errors map for a form to a field, just pass the `FieldError` object for that field. This lets us access the error safely via `error.address.zip` for example instead of unsafely as `errors[name]`


src/pdfFiller/1040FieldMappings.ts
```diff
@@ -0,0 +1,115 @@
...
+  11: (s) => s.taxPayer?.spouse?.ssid,
+  12: (s) => s.taxPayer?.primaryPerson?.address?.address,
+  13: (s) => s.taxPayer?.primaryPerson?.address?.aptNo,
+  14: (s) => s.taxPayer?.primaryPerson?.address?.city,
+  15: (s) => s.taxPayer?.primaryPerson?.address?.state,
+  16: (s) => s.taxPayer?.primaryPerson?.address?.zip,
+  17: (s) => s.taxPayer?.primaryPerson?.address?.foreignCountry,
+  18: (s) => s.taxPayer?.primaryPerson?.address?.province,
+  19: (s) => s.taxPayer?.primaryPerson?.address?.postalCode,
+  20: (s) => false,
```
zakpatterson, line 72: These and each of the commented lines below should become issues.

src/pdfFiller/fill1040Fields.js
```diff
@@ -1,57 +1,40 @@
...
+      fieldMappings[index](information)
+    ) {
+      pdfField.check()
+    } else if (pdfField instanceof PDFTextField) {
+      const defaultValue = index.toString()
+      let addValue
+      if (Object.prototype.hasOwnProperty.call(fieldMappings, index)) {
+        addValue = fieldMappings[index](information)
+      }
+      pdfField.setText(addValue ?? defaultValue)
```
zakpatterson, line 35: For now this is still leaving the field index numbers in fields with no content.

zakpatterson (commented):
Ok, this is a pretty big one, but I don't think there's a good way to minimize it. Let's see what you think.

zakpatterson:
I'm just going to leave it as multiple commits. When squashing we should narrow down the commit message.

src/components/Patterns.ts
```diff
@@ -25,7 +25,6 @@ export const Patterns: {[name: string]: PatternConfig} = {
     description: 'Input should be filled with 9 numbers'
   },
   currency: {
-    mask: '$999999',
```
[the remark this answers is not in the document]
thegrims: This should be mitigated by react-number-format once labelledInput is migrated

thegrims:
I think primary taxpayer info, spouse/dependant info, Family Contact Information, Filing Status should be added as 4 separate pages since the information is only saved on submit, with lots of form elements on the page the user may close the page before submitting and lose their work.

thegrims:
Dependent Information add button is spaced futher away from title text than Spouse Taxpayer Information is spaced away from it's title text in preview

thegrims:
Were the only options in filing status intended to be single and widower?

zakpatterson:
> Were the only options in filing status intended to be single and widower? Those are the only ones I'm seeing in the filing status dropdown

It's depends if there is a spouse or dependents.

thegrims:
There appears to have been a regression causing form fields / form widths to be different compared to what's in https://ustaxes.org/taxpayerinfo for example

thegrims:
image
The content here should all be contained inside the same card

thegrims:
Dependent, Spouse, and W2 add sections need form validations, ex not allowing empty fields from being submitted

thegrims:
What does this dot signify?
image

src/pdfFiller/1040FieldMappings.ts
thegrims: Thanks for commenting a description of each field index, that's really helpful!

zakpatterson:
> There appears to have been a regression causing form fields / form widths to be different compared to what's in https://ustaxes.org/taxpayerinfo for example

This is not the case. At present the width of the form / card container is determined by the width of the largest component. In the live site, the labels are longer, so the card is wider, and the textFields with fullWidth set to true are more noticeably longer.  We need to fix the styles for the form to have a larger minimum width rather than have their width be determined by their child elements.

Current deploy of this branch:

scrots_img_015

zakpatterson:
> There appears to have been a regression causing form fields / form widths to be different compared to what's in https://ustaxes.org/taxpayerinfo for example

I think the unmasked inputs are still longer, as they're set to `fullWidth` and expand to the width of the container, but now the containers are much narrower so the difference is almost not noticeable. So I'll try to figure that out.

zakpatterson:
Ok, I think this addresses your comments for the most part. The w2 job info right now still allows non-numeric input, and I'm not sure exactly why.

thegrims:
pressing save and continue on taxpayer info with empty form fields correctly stops the users from progressing, but doesn't display input is required errors

thegrims:
> > There appears to have been a regression causing form fields / form widths to be different compared to what's in https://ustaxes.org/taxpayerinfo for example
> 
> I think the unmasked inputs are still longer, as they're set to `fullWidth` and expand to the width of the container, but now the containers are much narrower so the difference is almost not noticeable. So I'll try to figure that out.

This is definitely because of the bad / shaky way I wrote the CSS for forms, That's definitely been on my list for things to look into improving in the future, but I'm no CSS expert. Thanks for fixing the widths on this PR!

thegrims:
> Ok, I think this addresses your comments for the most part. The w2 job info right now still allows non-numeric input, and I'm not sure exactly why.

Interesting, what have you been able to type in? I haven't been able to type non-numeric chars in the preview

zakpatterson:
It is just these fields now:

image

Also I just noticed another issue where the primary taxpayer doesn't show up in the employee drop down right away. 

I will have a little less time in the next few days. Also I'm dealing with a finger tendon laceration injury that required surgery so my output will be less for a bit!

thegrims:
Ouch! Take it easy, hope your finger heals well :)

zakpatterson:
Ok, I think I fixed up the error handling nicely. Give it another look when you can! Thanks.

thegrims:
Unit number should not be required in taxpayer information

thegrims:
This might be a possible solution to clearing local storage from netlify https://stackoverflow.com/a/57120853/5853454

thegrims:
image
Looks like the only employee option in job information is the spouse

thegrims:
Looks like the "save taxpayer information" action when saving "married filing jointly" is overwriting the primary taxpayer information in local storage and that's why only the spouse was showing up in W2 input
image

thegrims:
Looks like the save taxpayer info in contact information also does this

thegrims:
Filing status is smaller than other headers

thegrims:
Dependents are not filled into w2 form when create pdf button is pressed in preview

zakpatterson:
Is there a way to thread replies on here? Maybe if you add comments via review changes only, then we can address them in threads one by one instead of using quote replies.

zakpatterson:
Also noticing I dropped foreign country. Adding that.

src/components/TaxPayer/TaxPayer.tsx
```diff
@@ -0,0 +1,136 @@
...
+          label="City"
+          register={register}
+          name="address.city"
+          patternConfig={Patterns.name}
+          required={true}
+          error={errors.address?.city}
+          defaultValue={taxPayer?.primaryPerson?.address.city}
+        />
+        <LabeledCheckBox
+          label="Do you have a foreign address?"
```
thegrims, line 124: This label should be changed to "check the box below if you have a foreign  address since the options are no longer yes and no

src/pdfFiller/1040FieldMappings.ts
```diff
@@ -4,32 +4,27 @@ interface FieldMapping {
...
 
-const depField = (s: Information, fieldIdx: number): string | boolean | undefined => {
-  const fs = depFields(s, Math.floor((fieldIdx - 32) / 5))
-  if (fs !== undefined) {
-    return fs[(fieldIdx - 32) % 5]
+  if (depIdx < deps.length) {
+    const dep = deps[depIdx]
+    // Based on the PDF column, select the correct field
+    const depFieldIdx = (pdfFieldIdx - 32) % 5
+    return [dep.lastName, dep.ssid, dep.relationship, false, false][depFieldIdx]
```
thegrims, line 17: Looks like only the last name is getting added for dependents. Looking at the 1040 form, it should be first name concatenated to last name  with a space in between

zakpatterson: done.

src/components/TaxPayer/TaxPayer.tsx
zakpatterson: good?

zakpatterson: The alignment is not great, revising.

src/pdfFiller/1040FieldMappings.ts
```diff
@@ -0,0 +1,111 @@
...
+// so create field mappings for 4x5 grid of fields
+const depFieldMappings: FieldMapping = Object.fromEntries(
+  Array.from(Array(20)).map((u, n: number) => [n + 32, (s: Information) => depField(s, n + 32)])
+)
+
+const wages = (s: Information): number => {
+  if ((s.w2s ?? []).length < 1) {
+    return 0
+  }
+  return s.w2s.map((w2) => w2.income).reduce((l, r) => l + r)
```
thegrims, line 33: The wage strings are getting concatenated here rather than the incomes being summed

src/pdfFiller/1040FieldMappings.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+const fedWithholding = (s: Information): number => {
+  if ((s.w2s ?? []).length < 1) {
+    return 0
+  }
+  return s.w2s.map((w2) => w2.fedWithholding).reduce((l, r) => l + r)
```
thegrims, line 40: same for witholding

src/components/TaxPayer/TaxPayer.tsx
zakpatterson: check again

src/pdfFiller/1040FieldMappings.ts
thegrims: nice use of map reduce though!

src/components/TaxPayer/TaxPayer.tsx
thegrims: The green/grey switch is a cool concept, but I think it's a little distracting, should probably stick with black font for the checkbox instructions. Also, I was thinking about it and the yes/no checkbox text should probably be removed if the label is going to be _Check if you have a foreign address_  since that text combination is a little confusing.

src/pdfFiller/1040FieldMappings.ts
zakpatterson: Yep, all types are gone after compilation, so nothing stopping user input strings from going into fields assumed to be numbers.

src/pdfFiller/1040FieldMappings.ts
zakpatterson: Also resolved by sanitizing data before dispatching action.

src/components/TaxPayer/TaxPayer.tsx
zakpatterson: Ok, revised to single line, and took out the material ui styled title label thing.

thegrims:
Looks Great! Merging

PR #78, opened 2021-02-04
Computing 1040 and form stubs

zakpatterson:
Starting to lay the foundation for more schedules to be built and populated from our redux state


package.json
```diff
@@ -57,6 +57,7 @@
     ]
   },
   "devDependencies": {
+    "@types/node-fetch": "^2.5.8",
```
thegrims, line 60: Why is node fetch added to dev dependencies? It is already in dependencies

src/pdfFiller/fillPdf.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+  * Attempt to fill fields in a PDF from a Form,
+  * checking one by one that each pdf field and Form value
+  * Make sense by type (checkbox => boolean, textField => string / number)
+  */
+export function fillPDF (pdf: PDFDocument, form: Form): void {
+  const fields = form.fields()
+  const formFields = pdf.getForm().getFields()
+  // fill fields with fieldNumber
+  // check all boxes
```
thegrims, line 21: These comments aren't accurate any more, they can be removed

thegrims:
Looks great!

Issue #42, opened 2021-01-13
add precommit hook to fix linting errors when commiting


thegrims:
lint-staged and pretty-quick both look like really good options

https://prettier.io/docs/en/precommit.html

zakpatterson:
Can you just add a check for this in netlify? `yarn run lint` will return with an exit status code of 1 if there were any errors, 0 if there were no errors.

thegrims:
The idea is to add the hook not to detect errors but to have them be automatically & proactively fixed so a contributor can develop without worrying about formatting and then when committing, files are auto formatted to the projects formatting style

zakpatterson:
All hooks can be bypassed though so it's still needed in CI anyway imo.

thegrims:
@zakpatterson Changed the netlify build command to `CI=false npm run lint && npm run build`

thegrims:
`CI=false` makes warnings not fail the build

PR #81, opened 2021-02-06
Lint hook and minor fixes

thegrims:
- addresses #42 
- removes fill1040Fields.js which has been replaced by fillPdf.ts
- changes URL casing to all lower case from camel case to prevent user error when typing URL


thegrims:
@zakpatterson I don't think I can request a review from you but take a look if you're interested!

zakpatterson:
I tried this out and it disrupts my flow too much. I do 99% of my git actions via the Gitsavvy SublimeText plugin (similar in use to emacs' magit plugin and others), so commits and reordering / squashing commits are single or two keystrokes. I tried making several commits, then I did a commit, rebase, move up, squash with previous. All such actions trigger the linter to run again so it's too disruptive.

I think if we're going to have the husky hook then I would like it to block pushes only. Even if a commit chain A->B->X->D is pushed where only X fails linting, since code is only merged into master via squashed PRs, unlinted commits won't make it into mainline anyway. Does that make sense?

package.json
```diff
@@ -69,6 +69,16 @@
     "eslint-plugin-promise": "^4.2.1",
     "eslint-plugin-react": "^7.22.0",
     "eslint-plugin-standard": "^5.0.0",
+    "husky": "^4.3.8",
+    "lint-staged": "^10.5.4",
     "typescript-eslint": "0.0.1-alpha.0"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
```
zakpatterson: I think this should be 'pre-push' if that exists, what do you think

thegrims:
I changed the hook to pre-push. Let me know if that works for you :)

zakpatterson:
@thegrims I've tried to get this to trigger and it does not for me: https://github.com/zakpatterson/UsTaxes/commits/lintHook-test

Is it still trying to only lint staged files only?

```
➜  ustaxes git:(lintHook-test) git push
husky > pre-push (node v14.15.4)
ℹ No staged files found.        <<<<<-------*****
Enumerating objects: 9, done.
Counting objects: 100% (9/9), done.
Delta compression using up to 8 threads
Compressing objects: 100% (5/5), done.
Writing objects: 100% (5/5), 438 bytes | 438.00 KiB/s, done.
Total 5 (delta 4), reused 0 (delta 0)
To github:zakpatterson/ustaxes
   41c0a32..89ec4b6  lintHook-test -> lintHook-test
```

zakpatterson:
Yeah I see the `lint-staged` part. Since it's only happening on push, I think it's fine to just lint the entire project, or else I guess you could filter based on files changed since `origin/master`, which would assume the developer uses origin as the name for their remote:

```
➜  ustaxes git:(lintHook-test) git diff --name-only origin/master | cat
package-lock.json
package.json
src/components/Main.tsx
src/irsForms/F1040.ts
src/pdfFiller/fill1040Fields.js
➜  ustaxes git:(lintHook-test) 
➜  ustaxes git:(lintHook-test) git diff --name-only origin/master | cat
package-lock.json
package.json
src/components/Main.tsx
src/irsForms/F1040.ts
src/pdfFiller/fill1040Fields.js
➜  ustaxes git:(lintHook-test) 
```

thegrims:
@zakpatterson What are your thoughts on this husky hook?

`"husky": {
    "hooks": {
      "pre-push": "eslint ./src --ext .ts --ext .js --ext .tsx --ext .jsx"
    }
  },`

zakpatterson:
Maybe just reuse the `npm run lint` script

thegrims:
> Maybe just reuse the `npm run lint` script

The reason I didn't do that is wouldn't that not work with yarn?

zakpatterson:
Ah yea ok. Seems good then merge away.

zakpatterson:
Yep tried it out, seems to work well

PR #82, opened 2021-02-08
F1040 bugfixes and minor edits

thegrims:
- Set printed pdf fields to read-only
- Return W2 field rather than always being undefined
- Fix checkmark location for more than 4 children so checkmark is placed on more than 4 children box rather than Were born before January 2, 1956 box


thegrims:
@zakpatterson

PR #84, opened 2021-02-09
SVG wave

thegrims:
Added SVG wave to background of the form element.


zakpatterson:
Thats fun. The green color stops on the bottom if the window height is less than the form container height. Maybe that could be better.
2021-02-11-140220_1920x1080_scrot

Issue #68, opened 2021-02-02
Support 1099 interest and dividends


zakpatterson:
Lines 53-56 of the 1040, schedule b: https://www.irs.gov/pub/irs-pdf/f1040sb.pdf

Tracking todos:

* [x] #118 - Support income reported on 1099-INT, taxable interest 
* [x] #124 - Support dividends reported on 1099-DIV (Schedule B Part II) - This is complicated because qualified dividends need to be captured on the qualified dividends and capital gains worksheet, probably requires #69 first
* [ ] Support questions required in part 3 of Schedule B -- I think this should be a new issue, there are lots of informational questions
* [ ] Support tax-exempt interest? F1040 Line 2a

zakpatterson:
I think this is closed as a result of #118 and #124. Unchecked boxes above are among myriad other separate features yet to be considered.

PR #118, opened 2021-02-19
Add income user form supporting 1099-INT, add Schedule B filling

zakpatterson:
Add support for 1099-INT

This moves supported IRS tax forms into source. This may not be sustainable as we continue to add forms, states, etc. It's done this way now to avoid cross-origin request problems, and so the application works offline. We may want to in the future host the files separately.

This PR also has a side effect that the output PDF no longer has editable form fields. This is due to how I had to merge PDFs together. Each separate PDF schedule is filled, then serialized to bytes, then loaded into a new PDF and merged together. Not doing that intermediate step meant the fields were preserved but the filled data is lost.

Addresses the interest portion of #68. 

Does not handle yes no questions in part 3 of schedule b.


thegrims:
image
Needs a space after payer in 1099 Information

thegrims:
1099 label is still todo
image

zakpatterson:
@thegrims 
Fixed. 

scrots_img_017

src/irsForms/ScheduleB.ts
```diff
@@ -0,0 +1,88 @@
...
+      Array
+        .from(Array(14))
+        .flatMap((v, i) => {
+          if (i < ints.length) {
+            return [ints[i].payer, ints[i].income.toString()]
+          }
+          return ['', '']
+        })
+    )
+  }
```
thegrims, line 34: Minor nitpick, I think this could be more clear as
```
return (
      Array
        .from(Array(14))
        .flatMap((_, i: number) => {
          if (i < ints.length) {
            return [ints[i].payer, ints[i].income.toString()]
          }
          return ['', '']
        })
    )
```

zakpatterson: Yea that could definitely be better. [See if you like the new diff](https://github.com/thegrims/UsTaxes/pull/118/commits/0cdb4e6d0ba90927682822e49169a8b7df665ede#diff-03642ce9739ea7e819879e4cfb46c2e76a31045ee41b6dbc44945da7a8b1a4f8R32)

thegrims: That's much more clear, thank you!

PR #124, opened 2021-02-21
Add ScheduleD and handling 1099-B + 1099-DIV

### LINKED ISSUE #69: Support capital gains income (1099 B), Schedule D
Should be able to add data from multiple 1099 B forms, pdf output should include schedule D and accumulate to field number 64 on the 1040.

Todos:
* [x] [Schedule D or Capital Gains and Qualified Dividends worksheet](https://www.irs.gov/instructions/i1040sd#idm140312989632848)
* [x] Schedule D
* [x] Lines on 1040 (3a, 7)

Dealing with uncovered and unreported transactions is a separate issue:
* [ ] Schedule 8949 - used when basis is not reported to IRS, or for capital gains that are not reported on 1099B.
    * There are a number of types of transactions to report here:
        * (B) Short-term basis unreported transactions - required
        * (C) Short-term transactions not reported on form 1099-B - required
        * (E) Long-term basis unreported transactions - required
        * (F) Long-term transactions not reported on form 1099-B - required
    * User should be able to add list of transactions for B, C, E, F (note we should determine long term / short term)

zakpatterson:
Fills Schedule D correctly for covered gains / losses only (gains and losses reported on 1099-B with cost basis information reported to IRS. If cost basis information is not provided to IRS, then we need schedule 8949, which will be for later work

Computes tax on LTCG and qualified dividends according to the Qualified Dividends and Capital Gains worksheet (not line by line though)

Closes #69


src/redux/data.ts
```diff
@@ -57,17 +57,30 @@ export interface IncomeW2 {
 }
 
 export enum Income1099Type {
+  B = 'B',
```
thegrims, line 60: I think the strings here should be changed to 1099-B and 1099-INT for clarity

zakpatterson: Fixed this in the drop down code text mappign

thegrims:
image
I think the 'Enter name of bank, broker firm, or other payer' field and the 'Recipient' field should be moved before the 'Form Type' field. Right now it could be confused as associated with 'short term gains' but not 'long term gains'

src/components/income/F1099Info.tsx
```diff
@@ -4,25 +4,42 @@ import { Avatar, Box, Button, IconButton, List, ListItem, ListItemAvatar, ListIt
...
 
+const showIncome = (a: Supported1099): string => {
+  switch (a.type) {
+    case Income1099Type.INT: {
+      return a.form.income.toString()
+    }
+    case Income1099Type.B: {
+      const ltg = a.form.longTermProceeds - a.form.longTermCostBasis
+      const stg = a.form.shortTermProceeds - a.form.shortTermCostBasis
+      return `L(${ltg}), S(${stg})`
```
thegrims, line 20: Can this be changed to `Long term gains: ${ltg}, Short term gains: ${stg}` for clarity?

src/components/income/F1099Info.tsx
```diff
@@ -53,29 +70,72 @@ function List1099s (): ReactElement {
...
+        }
+      }
+    }
+    case Income1099Type.B: {
+      return {
+        payer: input.payer,
+        personRole: input.personRole,
+        type: input.formType,
+        form: {
+          shortTermCostBasis: parseInt(input.shortTermCostBasis),
```
thegrims, line 106: I think 2 decimal places should be supported for these fields because the final rounded integer profit of the transaction can vary based on the decimal values that are subtracted. Ex 2-1 is rounded to 1,  2.00 - 1.50 is rounded down to 0

Could use something like
`(Math.round(num * 100) / 100).toFixed(2);`

zakpatterson: Agree but the same problem exists for the W2 and remember we are using the same currency pattern everywhere:

```
currency: {
    regexp: /^[0-9]+$/,
    description: 'Input should be filled with numbers only'
  },
```

Which doesn't accept decimals. So I think this is a separate issue now.

src/components/income/F1099Info.tsx
zakpatterson: Yep, I need to figure out a better way to format multi line here. Agree it should be clearer.

src/components/income/F1099Info.tsx
thegrims: That makes sense. I think this issue would address the problem https://github.com/ustaxes/UsTaxes/issues/18

src/irsForms/ScheduleB.ts
```diff
@@ -46,11 +50,20 @@ export default class ScheduleB implements Form {
   )
 
   // TODO - 1099 DIV results
```
thegrims, line 52: should this TODO be removed?

src/data/federal.ts
```diff
@@ -0,0 +1,136 @@
...
+        exemptions: [
+          {
+            name: 'Standard Exemption (Single)',
+            amount: 0
+          }
+        ]
+      }
+    }
+  },
+  ltcg: {
```
thegrims, line 115: can this be changed to longTermCapitalGains or similar

src/redux/data.ts
```diff
@@ -57,17 +57,37 @@ export interface IncomeW2 {
...
-  INT = 'INT'
+  B = 'B',
+  INT = 'INT',
+  DIV = 'DIV'
 }
 
 export const form1099Types: Income1099Type[] = [
-  Income1099Type.INT
+  Income1099Type.B,
+  Income1099Type.INT,
```
thegrims, line 67: What is the advantage of having the 1099 form type strings in an enum vs putting them directly in the form1099Types array directly like

form1099Types: string[] = [
  'B',
  'INT',
  'DIV'
] 

or perhaps using the enum directly?

src/redux/data.ts
```diff
@@ -57,17 +57,37 @@ export interface IncomeW2 {
...
+export interface IntData {
   income: number
-  formType: Income1099Type
+}
+
+export interface DivData {
+  dividends: number
+  qualifiedDividends: number
+}
+
```
thegrims, line 86: I think these should be 1099BData, 1099IntData, 1099DivData for clarity

src/redux/data.ts
zakpatterson: Given an enum I don't think you can iterate over its keys without defining an array like this, is that not right?

src/redux/data.ts
zakpatterson: Done

src/redux/data.ts
thegrims: I think that you can get the values using `Object.values(MyEnum)` but that might decrease clarity

https://stackoverflow.com/questions/39372804/how-can-i-loop-through-enum-values-for-display-in-radio-buttons

src/components/income/F1099Info.tsx
zakpatterson: updated this

src/redux/data.ts
zakpatterson: Ah Object.values does work, doesn't need an explicit type cast. Good!

thegrims:
I like the INT, DIV, and B in the icon! Can the interest and dollar amounts be prefixed with a dollar sign like the 1099B form list item?
image

zakpatterson:
> I like the INT, DIV, and B in the icon! Can the interest and dollar amounts be prefixed with a dollar sign like the 1099B form list item?

Sure, added that. It would be nice to have an ability to enforce consistent formatting across the UI, after #18.

src/util.ts
```diff
@@ -0,0 +1,19 @@
+
+export const anArrayOf = <A>(n: number, a: A): A[] => Array.from(Array(n)).map(() => a)
+
+/**
+ * Given two arrays [a1, a2, ...], [b1, b2, ...] make an array
```
thegrims, line 5: This util content should be moved into the util.ts file in `src\irsForms` rather than being in the `src` directory.

zakpatterson: I put them here because they are so generic that they aren't specific or related to what our project is doing. Whereas the functions in irsforms are specific to displaying data in those forms.

zakpatterson:
yay!

PR #131, opened 2021-02-25
add default 0 for reduce if there is no first value in array

thegrims:
* fix for if user attempts to create tax pdf without dividends
* fix for if user attempts to create tax pdf without w2 income


thegrims:
Merging to fix bug

PR #132, opened 2021-02-25
Render numeric fields using number-format

### LINKED ISSUE #18: Investigate using react-number-format over react-input-mask

zakpatterson:
Fixes #18  

This allows us to continue to use LabeledInput for any simple text field, but inject the logic of formatting as any kind of number just as before. The difficulty is that numeric fields now require the form-hook's control property, whereas they did not previously. So there is a little extra syntax to require the control property for numeric fields and not for text fields.


src/components/RefundBankAccount.tsx
```diff
@@ -53,7 +53,7 @@ export default function RefundBankAccount (): ReactElement {
                 label="Bank Routing number"
                 register={register}
                 required={true}
-                patternConfig={Patterns.bankRouting}
+                patternConfig={Patterns.bankRouting.get(control)}
```
zakpatterson, line 56: This is the main refactoring change. 

It might be better to just split numeric fields and text fields into separate components, since numeric fields now require access to the control object and text fields do not. 

On the other hand, I like having all the field configurations set up concisely in Patterns, then having one component handle the rendering logic.

thegrims:
I was able to submit from 'Taxpayer Information' with a zip that did not consist of 9 numbers

zakpatterson:
> I was able to submit from 'Taxpayer Information' with a zip that did not consist of 9 numbers

Think I got that.

thegrims (commented):
image
Bank account number not validating correctly

zakpatterson:
Got that, nice catching...

thegrims (approved):
LGTM

PR #138, opened 2021-03-04
Add contributor oriented documentation

### LINKED ISSUE #135: Add ARCHITECTURE.md describing high-level structure of the project
Please see this article - https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html

This will be super super useful for contributors to quickly jump into contributing.

zakpatterson:
Also fixes up readme language. I removed the caution note about not using
this software for the 20/21 tax season. The reason is just that some
people file extensions and then don't complete their returns until
October. I just don't think there's much reason to be concerned
that someone will trust us to the point of just blindly printing out
their return and mailing it in.

Additionally, we'll probably never be 100% confident in our output,
even after years of work we can still make mistakes that would be
created or fixed by other people's contributions.

Fixes #135


README.md
```diff
@@ -28,24 +28,27 @@ Users who only have wage income and live in the below states should be able to f
...
-- If you wish to make a code contribution please check out current issues and comment `@thegrims` on the issue that you want to work on so that multiple contributors don't end up working on the same issue. 
+- If you wish to make a code contribution please check out current issues and leave a comment so that multiple contributors don't end up working on the same issue.
 
 - If you have an idea for an improvement or a bugfix, feel free to file an issue to be reviewed.
 
 # Running
 This application can be either run as a web application or a standalone desktop application. 
 
-### Web application
+### Web application4
```
thegrims, line 46: This looks like a typo

zakpatterson:
Very rough at this point. Hoping for discussion.

ARCHITECTURE.md
```diff
@@ -0,0 +1,60 @@
...
+
+* This project uses typescript and is built with [NPM](https://www.npmjs.com/get-npm)
+
+### Desktop application requirements
+
+* The desktop application is built with [Tauri](https://tauri.studio/). While this project does not have any significant rust code yet, this may be built out in the future. Running the application as a desktop app requires [Rust](https://www.rust-lang.org/) installed.
+
+## Project design
+
+There are two main concerns separated in this project. Data must be collected from users and that data must be rendered into a PDF file when the user exports their 1040 and attachments.
```
thegrims, line 14: In my mind, there are 4 main logical components
* user input pages
* redux storage functionality actions & interfaces
* tax calculations
* PDF form filling

ARCHITECTURE.md
```diff
@@ -0,0 +1,60 @@
...
+
+In order to separate these concerns, the data collected from the webforms is stored in a single data object, with schema defined in `src/redux/data.ts`: 
+
+As of now, the crucial root schema is:
```
thegrims, line 18: Maybe could change to 

_The root schema of stored data from form submissions is:_

or similar

ARCHITECTURE.md
zakpatterson: Yea important to discuss this part. So far the pdf form filling drives the tax calculations really. For example ach line of the 1040 and other schedules are implemented as separate functions that depend on the output of all the previous lines. 

The only part where we deviate is in the Capital Gains and Qualified Dividends Worksheet which is laborious and designed for someone with a pencil and paper only.

zakpatterson: I like this four bullet explanation as you've laid it out, I added a drawing about that.

thegrims:
Not a huge issue but in dark mode the diagram is somewhat hard to pick out. Maybe for black text a white background should be used rather than transparent. A different diagram color would work too

image

thegrims:
Should link to the Tauri environment setup guide in the desktop application requirements section  https://tauri.studio/en/docs/getting-started/intro#setting-up-your-environment

thegrims:
I'm not sure if this would overcomplicate the diagram but you also might want to add in rehydration from localstorage

ARCHITECTURE.md
thegrims: > So far the pdf form filling drives the tax calculations really

For sure. The reason I mentioned tax calculations and PDF forms as separate is I think it will be good to keep these areas encapsulated from each other for example if the PDF form fields move around but the calculations stay the same or if we want to add an expected tax preview at each step in or similar in the future.

zakpatterson:
> I'm not sure if this would overcomplicate the diagram but you also might want to add in rehydration from localstorage

I think this is 

> I'm not sure if this would overcomplicate the diagram but you also might want to add in rehydration from localstorage

I just added it as a note under the diagram because it happens automatically and it's not something a contributor has to modify if they're adding a new component.

ARCHITECTURE.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+* **f1099s**: An array of all 1099s that have been added. Note this includes 1099B which goes to Schedule D, 1099-Int which goes to schedule B, and 1099-DIV which provides data that goes to both Schedule B and Schedule D. This confusion is not needed at this level of the data model. Later when PDFs are created, the correct data can be accessed by the code managing those schedules.
+* **w2s**: All W2s that have been added for both primary taxpayer and spouse
+* **refund**: Direct deposit information
+* **taxPayer**: Basic information about user's name, ssn, dependents, spouse
+
+### PDF Export
+
+Supported tax forms are included in the source control of this repository. [src/irsForms/](src/irsForms/) includes all Schedules that can be filled by this project. Each of these Schedule definitions implements this interface:
```
thegrims, line 48: schedules shouldn't be capitalized

thegrims (approved):

PR #142, opened 2021-03-09
Start a reading list to build code culture; a domain specific reading…

breck7:
… list; and a list of links to relevant sites and projects


thegrims:
While the reading list and related links are interesting, I think the readme should contain strictly information related directly to the project.

PR #143, opened 2021-03-09
Suggested rule of thumb: every folder should have a readme. A little …

breck7:
…documentation today is better than a lot of documentation tomorrow, and it keeps a project as flat as logically possible.


public/readme.md
```diff
@@ -0,0 +1,7 @@
+# Public
+
+This folder contains static assets served by the web server.
+
+`react-scripts` is used to serve the site during development.
```
zakpatterson, line 5: I think this goes to just what the structure of an application created using create-react-app looks like. Can this be in the root readme?

public/forms/readme.md
```diff
@@ -0,0 +1,13 @@
+# Forms
+
+This folder contains the official PDFs published by the IRS.
+
+[todo: why do we have some forms and not all of them?]
```
thegrims, line 5: Right now only forms that are being filled by the application are in the forms directory

zakpatterson:
Going to close this for now. If you would like to revisit it and add more documentation that would be helpful!

PR #145, opened 2021-03-10
Add release badge

zakpatterson:
using badge from badgen.net


thegrims:
Minor nitpick but I think the badge should go next to the netlify badge since they are both related to build status

PR #148, opened 2021-03-14
Refactor pager hook and use context

zakpatterson:
By using context instead of props we no longer have to pass the same props to every single page component in our routing. Also because each page can retrieve the needed pager instructions via context, the pages are all propless, and the routes can be created more simply by just mapping over a data structure containing info about all our routes.

Before, writing our list of routes was tedious, as for example:

```ts
        <Route path="/createpdf" exact>
          <CreatePDF navButtons={allDoneButtons} />
        </Route>
```
Requires the final page set of buttons passed. Every page needs to get these buttons passed explicitly. 

This is solved by using a react context and allowing the Main component to set the appropriate buttons for the context.

This refactoring involves wrapping all the pages in a `<PagerContext.Consumer>`. This messes up the diffs due to indenting in the rendering for each page's component.


src/components/Main.tsx
```diff
@@ -46,77 +46,61 @@ const Urls = {
...
+      item('Wages (W2)', Urls.income.w2s, <W2JobInfo />),
+      item('Income (1099)', Urls.income.f1099s, <F1099Info />)
     ]
   },
   {
     title: 'Results',
     items: [
-      ['Refund Information', Urls.refund],
-      ['Review and Print', Urls.createPdf]
+      item('Refund Information', Urls.refund, <RefundBankAccount />),
```
zakpatterson, line 74: Now page title, url, and page component are all in one place.

src/components/Main.tsx
```diff
@@ -46,77 +46,61 @@ const Urls = {
...
 
 const drawerSections: Section[] = [
   {
     title: 'Personal',
     items: [
-      ['Taxpayer Information', Urls.taxPayer.info],
-      ['Spouse and Dependents', Urls.taxPayer.spouseAndDependent],
-      ['Filing Status', Urls.taxPayer.filingStatus],
-      ['Contact Information', Urls.taxPayer.contactInfo]
+      item('Taxpayer Information', Urls.taxPayer.info, <TaxPayerInfo />),
```
zakpatterson, line 58: Now page title, url, and rendering component are all in one place

src/components/Main.tsx
```diff
@@ -46,77 +46,61 @@ const Urls = {
...
-          <W2JobInfo onAdvance={forward} navButtons={stepDoneButtons} />
-        </Route>
-        <Route path={Urls.income.f1099s} exact>
-          <F1099Info onAdvance={forward} navButtons={stepDoneButtons} />
-        </Route>
-        <Route path="/createpdf" exact>
-          <CreatePDF navButtons={allDoneButtons} />
-        </Route>
+        <Redirect path="/" to={Urls.default} exact />
+        <PagerContext.Provider value={{ onAdvance: (onAdvance ?? (() => {})), navButtons }}>
```
zakpatterson, line 97: This wraps all our pages now, providing the required buttons and next function.

src/components/RefundBankAccount.tsx
```diff
@@ -22,7 +22,7 @@ const toRefund = (formData: UserRefundForm): Refund => ({
   accountType: formData.isChecking ? AccountType.checking : AccountType.savings
 })
 
-export default function RefundBankAccount ({ navButtons, onAdvance }: PagedFormProps): ReactElement {
```
zakpatterson, line 25: Props are no longer needed as these are provided by context.

src/components/Main.tsx
```diff
@@ -46,77 +46,61 @@ const Urls = {
...
-  drawerSections
-    .flatMap((section: Section) => section.items)
-    .map((item) => item[1])
-)
-
 export default function Main (): ReactElement {
-  const [, forward, prevUrl] = usePager(allUrls)
+  const allItems: SectionItem[] = drawerSections.flatMap((section: Section) => section.items)
+
+  const [, onAdvance, prev] = usePager(allItems, (item) => item.url)
```
thegrims, line 83: any reason for the leading comma in 
`const [, onAdvance, prev]`

zakpatterson: If you're asking about what that leading comma does, it's just selecting the middle two elements from the array returned by that usePager function. I thought the pager hook would return other info like the current page. But this is the only use of it so I simplified it in 9bbb11b.

thegrims: That makes sense, thanks for the context

src/components/TaxPayer/TaxPayer.tsx
```diff
@@ -95,57 +95,58 @@ export default function TaxPayerInfo ({ navButtons, onAdvance }: PagedFormProps)
...
-        />
-        <LabeledCheckBox
-          label="Check if you have a foreign address"
-          control={control}
-          value={isForeignCountry}
-          setValue={updateForeignCountry}
-          name="isForeignCountry"
-        />
-        {csz}
+    <PagerContext.Consumer>{({ navButtons, onAdvance }) =>
```
thegrims, line 101: Can the pagercontext lines here be separated like below
```
<PagerContext.Consumer>
      {({ onAdvance, navButtons }) =>
          // somecode
      }
</PagerContext.Consumer>
```

thegrims: same in createPDF, F1099Info, W2JobInfo

zakpatterson: For sure, I will do that. The reason I didn't do it before was to try to minimize the diff. I didn't want to have to indent the entire block. But final code quality is more important. Will do.

src/components/pager.tsx
```diff
@@ -8,14 +8,15 @@ import { Link, useHistory } from 'react-router-dom'
...
+ * @param pages a list of pages A
+ * @param url gets a url out of a page A, starting with '/' (history.location.pathname)
+ * @returns [previousPage, goToNextPage]
  */
-export const usePager = (pageUrls: string[]): [() => void, () => void, string | undefined, string] => {
+export const usePager = <A, >(pages: A[], url: (a: A) => string): [A | undefined, (() => void) | undefined] => {
   const history = useHistory()
 
   const navPage = (path: string): number | undefined => {
-    const found = pageUrls.findIndex(p => p === path)
```
thegrims, line 18: Doesn't have to be done in this pull request, but maybe the PageUrls could be changed to a dictionary with the path to index mapping so that the list of strings is not looped through every time. Something like

`pageIndex = pageUrls[path]`

That way also this navPage function would not be needed.

thegrims:
When I hit save and continue on the first page in the preview the pager element progresses, but the rest of the page does not
image
image

src/components/pager.tsx
zakpatterson: I'll think about it. Dictionary key order is not maintained though, so I would have to use some other data structure that combines array indexing with hash lookup.

zakpatterson:
> When I hit save and continue on the first page in the preview the pager element progresses, but the rest of the page does not

good catch, fixed that.

src/components/TaxPayer/TaxPayer.tsx
zakpatterson: did that.

src/components/pager.tsx
thegrims: I was thinking the corresponding value of the path key would be the array index

thegrims (approved):
LGTM

src/components/pager.tsx
zakpatterson: Take one more look, I went ahead and found an optimization. Just using Map didn't quite work - it remembers the order of the keys but only gives an iterator. It's not indexable by key insertion order, only iterable.

thegrims: Looks great! Thank you

thegrims (approved):

PR #162, opened 2021-03-21
Adding primary and spouse options to be set as a dependent

thegrims:
- Increases top spacing on LabeledCheckBox
- Adds @types/react to resolve IDE error messages
- Fills in these two checkboxes in the 1040 form
image


src/redux/data.ts
```diff
@@ -29,6 +29,11 @@ export interface Address {
 
 export interface PrimaryPerson extends Person {
   address: Address
+  isTaxpayerDependent: boolean
+}
+
+export interface Spouse extends Person {
```
zakpatterson, line 35: This is a new interface. Check out the end of redux/validate.ts. 

I have found that is necessary to resolve the error you hit.

thegrims:
Looks like the standard deduction has to be modified for dependents as well

> Dependents – If you can be claimed as a dependent by another taxpayer, your standard deduction for 2020 is limited to the greater of: (1) $1,100, or (2) your earned income plus $350 (but the total can't be more than the basic standard deduction for your filing status).

https://www.irs.gov/taxtopics/tc551#:~:text=Dependents%20%E2%80%93%20If%20you%20can%20be,deduction%20for%20your%20filing%20status).

thegrims:
Incorporated logic from this standard deduction worksheet https://apps.irs.gov/app/vita/content/globalmedia/standard_deduction_chart_65_or_older_4012.pdf

src/components/TaxPayer/TaxPayer.tsx
```diff
@@ -113,6 +119,13 @@ export default function TaxPayerInfo (): ReactElement {
               defaults={taxPayer?.primaryPerson}
               control={control}
             />
+            <LabeledCheckBox
+              label="Check if you are a dependent"
```
zakpatterson, line 123: Could this be more clear?

src/components/TaxPayer/SpouseAndDependent.tsx
```diff
@@ -97,7 +105,15 @@ export const SpouseInfo = (): ReactElement => {
           register={register}
           errors={errors}
           control={control}
-        />
+        >
+          <LabeledCheckBox
+            label="Check if your spouse is a dependent"
```
zakpatterson, line 110: Maybe Can someone else claim spouse as a dependent?

src/irsForms/F1040.ts
```diff
@@ -169,6 +169,11 @@ export default class F1040 implements Form {
   standardDeduction = (): number => {
     if (this.filingStatus === undefined) {
       return 12400
+    } else if (this.isTaxpayerDependent || this.isSpouseDependent) {
+      return Math.min(
+        (F1040.standardDeductions[this.filingStatus] ?? 12400),
```
zakpatterson, line 174: I think all these ###s should be parameters in the federal.ts file. It's painful to imagine them changing year to year, but in general looking them up somewhere feels better.

zakpatterson (changes requested):
I tried this out and the checkboxes and form fields were what I expected in the output PDF, but I didn't scrutinize the numbers carefully. Looks good.

zakpatterson:
I'm a little confused about what happens when a taxpayer can be someone's dependent:

Page 32 of the [general instructions](https://www.irs.gov/pub/irs-pdf/i1040gi.pdf) shows the worksheet for the standard deduction:

scrots_img_023

So for these people, the standard deduction is supposed to be at least $1,100, and at most the lesser of their earned income + 350 or the normal standard deduction. So it appears to me they're supposed to lose the amount of the standard deduction that would otherwise apply to their unearned income.

At the moment if I try someone that has a 7000 W2, they are still getting their full $12,400 deduction

thegrims:
> I'm a little confused about what happens when a taxpayer can be someone's dependent:
> 
> Page 32 of the [general instructions](https://www.irs.gov/pub/irs-pdf/i1040gi.pdf) shows the worksheet for the standard deduction:
> 
> scrots_img_023
> 
> So for these people, the standard deduction is supposed to be at least $1,100, and at most the lesser of their earned income + 350 or the normal standard deduction. So it appears to me they're supposed to lose the amount of the standard deduction over $350 that would otherwise apply to their unearned income.
> 
> At the moment if I try someone that has a 7000 W2, they are still getting their full $12,400 deduction

Is your local branch caught up to the head of this branch or filingStatus undefined? When I added someone with an income of 7000, the deduction showed up as 7350.


```
if (this.filingStatus === undefined) {
      return 12400
    } else if (this.isTaxpayerDependent || this.isSpouseDependent) {
      return Math.min(
        (F1040.standardDeductions[this.filingStatus] ?? 12400),
        (this.wages() > 750) ? (this.wages() + 350) : 1100
      )
    }
    return F1040.standardDeductions[this.filingStatus]
```

thegrims:
Also a side note but should we remove the filing status undefined option? It seems like it should always be defined

src/irsForms/F1040.ts
thegrims: Good idea, that will help keep the code clean and easy to debug

zakpatterson:
I had left the filing status undefined, it works as you say. The snippet you pasted needs fixing. As it is now, if filing status is undefined then the standard deduction should be undefined.

If we get rid of undefined does that mean we have to choose for the user in some cases? If the user adds a spouse, then selects MFJ, then removes the spouse, should we switch the filing status to HoH or S depending if there is a dependent? I think we just show an unfinished task if the user tries to generate a return using an undefined filing status.

zakpatterson (approved):

zakpatterson:
Proposing a way to handle unselected filing status in #222

PR #178, opened 2021-03-24
Edit spouse details

jswiss:
This PR is a work in progress and will be split into multiple pull requests. The first enables users to edit spouse details. That logic is within, but remains to be tested. Future PRs will tackle editing functionality for dependents and forms.


thegrims:
The package-lock changes look like a pre npm 7 version of npm might have been used to install. What version of npm do you have on your machine?

jswiss:
> The package-lock changes look like a pre npm 7 version of npm might have been used to install. What version of npm do you have on your machine?

It was pre-npm 7; good catch. I'll update and re-install

zakpatterson:
The spouse edit button looks great. 

The dependent edit button does not seem to work.

src/components/TaxPayer/PersonFields.tsx
```diff
@@ -15,36 +16,40 @@ interface PersonFieldsProps<T extends Person> extends BaseFormProps {
...
     <LabeledInput
       label="First Name and Initial"
+      aria-label="First Name and Initial"
       register={register}
       name="firstName"
       patternConfig={Patterns.name}
       required={true}
       error={errors.firstName}
-      defaultValue={defaults?.firstName}
+      defaultValue={(person !== null) ? person?.firstName : defaults?.firstName}
```
zakpatterson, line 32: the defaultValue property is allowed to be undefined. I don't think there's a reason for a three case value here.

I think it could just be: `person?.firstName ?? defaults?.firstname`, as the `??` operator allows the right side expression if the left side expression is null or undefined.

jswiss: Good suggestion! I haven't used the nullish coalescing operator before, glad to get to put it to use.

src/components/TaxPayer/PersonFields.tsx
```diff
@@ -54,9 +59,10 @@ export const PersonFields = <T extends Person>({ register, control, errors, defa
 interface PersonListItemProps {
   person: Person
   remove: () => void
+  edit?: () => void
```
jswiss, line 62: I'll make this required once doing the dependents PR

src/components/TaxPayer/PersonFields.tsx
```diff
@@ -65,6 +71,11 @@ export const PersonListItem = ({ person, remove }: PersonListItemProps): ReactEl
       primary={`${person.firstName} ${person.lastName}`}
       secondary={person.ssid}
     />
+    <ListItemIcon>
+      <IconButton onClick={edit} edge="end" aria-label="edit">
+        <EditIcon />
+      </IconButton>
+    </ListItemIcon>
```
jswiss, line 78: You can only have one `ListItemSecondaryAction` child  element at a time.

zakpatterson:
Up to you if the dependent form editing is done in this PR or a subsequent one. If it's subsequent then just remove the unconnected edit button from the dependent form for this PR then

src/components/TaxPayer/SpouseAndDependent.tsx
```diff
@@ -75,44 +78,67 @@ export const AddDependentForm = (): ReactElement => {
 
 export const SpouseInfo = (): ReactElement => {
   const { register, control, errors, handleSubmit, getValues } = useForm<UserPersonForm>()
-  const [editSpouse, updateEditSpouse] = useState(false)
+  const [editSpouse, setEditSpouse] = useState<boolean>(false)
+  const [createSpouse, setCreateSpouse] = useState<boolean>(false)
```
jswiss, line 80: It might be nice to abstract these (and other?) hooks into custom hooks

src/components/TaxPayer/PersonFields.tsx
```diff
@@ -15,37 +16,41 @@ interface PersonFieldsProps<T extends Person> extends BaseFormProps {
...
+  person?: Person | null | undefined
   control: Control
 }
 
-export const PersonFields = <T extends Person>({ register, control, errors, defaults, children }: PersonFieldsProps<T>): ReactElement => (
+export const PersonFields = <T extends Person>({ register, control, errors, defaults, children, person }: PersonFieldsProps<T>): ReactElement => (
   <div>
     <LabeledInput
       label="First Name and Initial"
+      aria-label="First Name and Initial"
```
jswiss, line 27: Material UI components bring accessibility out of the box, but our custom elements don't always have the same attributes. I chucked in `aria-label` here, but it'd be worth raising a separate issue to do a full accessibility audit. 

Lighthouse gave us an 87, but it could be better 😄

src/components/TaxPayer/SpouseAndDependent.test.tsx
```diff
@@ -0,0 +1,250 @@
+import React from 'react'
+import { render, screen, fireEvent, waitFor } from '@testing-library/react'
+import { Provider } from 'react-redux'
+import userEvent from '@testing-library/user-event'
```
jswiss, line 4: I commented throughout the tests, hope they're easy to follow. Most of my approach came from here: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library

src/components/TaxPayer/SpouseAndDependent.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+import { SpouseInfo } from './SpouseAndDependent'
+import { store } from '../../redux/store'
+
+afterEach(async () => {
+  await waitFor(() => localStorage.clear())
+  jest.resetAllMocks()
```
jswiss, line 11: This is not working as expected. I need to properly mock `redux-persist`. Depending on the testing strategy moving forward, I can work on that next.

src/components/TaxPayer/SpouseAndDependent.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+})
+
+describe('SpouseInfo', () => {
+  it('renders an `Add` button when no spouse has been added', () => {
+    render(
+    <Provider store={store}>
+      <SpouseInfo />
+    </Provider>
```
jswiss, line 19: I'd like to put together a helper render function that takes in a Provider to abstract some of this away.

src/components/TaxPayer/SpouseAndDependent.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+    })
+
+    // get all three text inputs
+    const inputs = screen.getAllByRole('textbox')
+    const [firstNameInput, lastNameInput, ssidInput] = inputs
+
+    // add values for each input
+    userEvent.type(firstNameInput, 'Sally K')
+    userEvent.type(lastNameInput, 'Ride')
+    fireEvent.change(ssidInput, { target: { value: '123456789' } })
```
jswiss, line 90: I'm not sure why, probably due to nested elements, but `userEvent.type` was not working for ssid, so I opted for `fireEvent` instead.

src/components/TaxPayer/SpouseAndDependent.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+    }))
+
+    // expect the new names to be concatenated and new ssid to appear without hyphens
+    await screen.findByText('Fella McGee')
+    screen.getByText('987654321')
+
+    const deleteButton = screen.getByLabelText('delete')
+
+    // click the delete button
+    fireEvent.click(deleteButton)
```
jswiss, line 137: This bit down here is where I realised that mocking localStorage in `setupTests` is not enough; `redux-persist` needs to be mocked as well, as the data stored in localStorage is not being wiped and state is persisting between tests as a result

src/setupTests.js
```diff
@@ -1,5 +1,13 @@
...
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
 import '@testing-library/jest-dom/extend-expect'
+
+const localStorageMock = {
+  getItem: jest.fn(),
+  setItem: jest.fn(),
+  clear: jest.fn()
+}
+global.localStorage = localStorageMock
```
jswiss, line 13: This is the jest-approved way of mocking localStorage globally. Unfortunately this alone doesn't solve the redux persist issue.

src/components/TaxPayer/PersonFields.tsx
```diff
@@ -65,6 +71,11 @@ export const PersonListItem = ({ person, remove }: PersonListItemProps): ReactEl
       primary={`${person.firstName} ${person.lastName}`}
       secondary={person.ssid}
     />
+    <ListItemIcon>
+      <IconButton onClick={edit} edge="end" aria-label="edit">
```
zakpatterson, line 75: If it's undefined let's not display the edit icon for now, so the edit button is not visible for the dependents?

src/components/TaxPayer/SpouseAndDependent.test.tsx
zakpatterson: Is there a fixture that could be used? I looked around for a minute and it appears there are fixture libraries that would let you reset local storage before / after each test, that might work?

src/components/TaxPayer/SpouseAndDependent.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+    })
+
+    const inputsAfterDelete = screen.queryAllByRole('textbox')
+
+    // expect input fields to not be in the document
+    expect(inputsAfterDelete).toHaveLength(0)
+  })
+  it('does not save when required fields not completed', async () => {
+    render(
+    <Provider store={store}>
```
zakpatterson, line 151: this indentation seems off, not a big deal but there several instances.

src/components/TaxPayer/SpouseAndDependent.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+    // only the ssid error remains
+    await screen.findByText('Input is required')
+
+    // incorrectly enter ssid
+    fireEvent.change(ssidInput, { target: { value: '123sc' } })
+    fireEvent.click(createButton)
+
+    // expect ssid error to remain
+    await waitFor(() => {})
+    await screen.findByText('Input should be filled with 9 digits')
```
zakpatterson, line 223: can these find by text strings be sourced from code?

zakpatterson (changes requested):
looks like this is really close, i'm excited to get it in.

src/components/TaxPayer/PersonFields.tsx
```diff
@@ -15,37 +16,41 @@ interface PersonFieldsProps<T extends Person> extends BaseFormProps {
   defaults?: T
   children?: ReactNode
   errors: DeepMap<Partial<Person>, FieldError>
+  person?: Person | null | undefined
```
thegrims, line 19: This should be able to be simply set as `person?: Person` since the question mark denotes that the person interface property is possibly undefined.  From looking at the code it doesn't look like person needs to be able to be set to null, only undefined.

package-lock.json
```diff
@@ -2020,7 +2020,7 @@
       "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-0.14.0.tgz",
```
thegrims, line 2020: Do you know what the cause of these changes to the package-lock is?

thegrims:
The tests look great! Can the console.log and console.error statements generated by running the tests be silenced by adding something like the below to `src\setupTests.js` or similar?
```
global.console = {
  log: jest.fn(), 
  error: jest.fn(),

  // Keep native behaviour for other methods, use those to print out things in your own tests
  warn: console.warn,
  info: console.info,
  debug: console.debug,
};
```

jswiss:
> Up to you if the dependent form editing is done in this PR or a subsequent one. If it's subsequent then just remove the unconnected edit button from the dependent form for this PR then

Yeah, fair point. I'll just crack on with this then; I'll end up putting in a bunch of logic I'd have to strip out later, and run the risk of forgetting to do so, if I don't tackle dependents now

package-lock.json
jswiss: I had to run `--legacy-peer-deps` with npm to get packages installing properly on v7. Given that I haven't installed anything new, I shouldn't have committed this anyways, and will remove

jswiss:
> The tests look great! Can the console.log and console.error statements generated by running the tests be silenced by adding something like the below to `src\setupTests.js` or similar?
> 
> ```
> global.console = {
>   log: jest.fn(), 
>   error: jest.fn(),
> 
>   // Keep native behaviour for other methods, use those to print out things in your own tests
>   warn: console.warn,
>   info: console.info,
>   debug: console.debug,
> };
> ```

Yep, absolutely, will add now

src/setupTests.js
```diff
@@ -11,3 +11,13 @@ const localStorageMock = {
   clear: jest.fn()
 }
 global.localStorage = localStorageMock
+
+global.console = {
+  log: jest.fn(),
+  error: jest.fn(),
```
jswiss, line 17: I guess it's worthwhile keeping `warn`s in tests for now. We can disable those too if they prove too noisy

src/components/TaxPayer/PersonFields.tsx
zakpatterson: From this doc:

> The aria-label attribute is used to define a string that labels the current element. Use it in cases where a text label is not visible on the screen. If there is visible text labeling the element, use aria-labelledby instead. 

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute

Why add aria-label?

There probably is an accessibility issue though, because the label in `<LabeledInput>` is just in a separate `p` tag floating above the input:

```
      <Box display="flex" justifyContent="flex-start">
        <p><strong>{strongLabel}</strong>{label}</p>
      </Box>
      <Box display="flex" justifyContent="flex-start">
        {input}
      </Box>
```
That's probably not right. Should be a separate issue though.

zakpatterson:
Hey @jswiss, attempting to assist here since I caused some conflicts. Added PR to your fork, so if that approach makes sense then merge on your fork to include that merge and commit in this PR. Thanks!

src/components/TaxPayer/SpouseAndDependent.test.tsx
jswiss: Maybe. I think it's more about mocking the redux store and being able to clear the persisted cache. I tried a mocked store, which initially worked fine, but wouldn't update beyond the default state.

IMO it's probably worth parking for now until a proper testing strategy is defined; maybe we don't want to write 2:1 tests:code like I did here 😅

src/components/TaxPayer/SpouseAndDependent.test.tsx
jswiss: Yeah, I don't know what's happening with my editor throwing that off. I think once this is in I'd like to raise a separate PR to add a `prettier` step to the pipeline so mal-formatted code will never reach GitHub

zakpatterson (changes requested):
This looks great! I ran into a couple issues, which I proposed fixes for here:

https://github.com/jswiss/UsTaxes/pull/2

I hope we can find some kind of refactoring to make writing the tests easier.

src/components/TaxPayer/SpouseAndDependent.tsx
thegrims: Yeah that would b nice to have, not sure of a way to abstract further though

thegrims (approved):

zakpatterson (approved):

PR #192, opened 2021-03-30
Fix phone and zip patterns, and labeled input regression

zakpatterson:
#132 added a regression where LabeledInputs got a name pattern by default.

* Zip: Should accept 5 or 9 digit zips. Zip+4 is not mandatory
* Phone: Should not accept phone numbers starting with 0 or 1
* Email: Regression caused labeled input to default to name pattern, which meant email addresses could not contain numbers or even @.


thegrims (approved):
LGTM

PR #193, opened 2021-03-30
moving type dependencies to devDependencies

thegrims:
types should be moved to devDependencies since all type information will be removed during the build process transpilation from TS -> JS


package.json
```diff
@@ -8,13 +8,6 @@
     "@testing-library/jest-dom": "^5.11.10",
```
zakpatterson, line 8: this stuff too?

thegrims: Good idea!

zakpatterson (approved):

PR #206, opened 2021-04-04
Add docker setup

lekha:
**Purpose:**

Standardize local development environment among contributors and speed up time for new contributors to start coding.

For now, Docker is only available for the browser. The desktop application will need additional work before it can be used through Docker too. Ideally, we will also eventually use this setup for production deployment.

**Testing:**

```
docker-compose build
docker-compose up
```
Open a browser to `http://localhost:3000` and the website opens!

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,13 @@ This application can be either run as a web application or a standalone desktop 
 
 To run, `npm start` or `yarn start`.
 
+If preferred, a Docker alternative is available:
+```
+docker-compose build
+docker-compose up
+```
+Then, open a browser to `http://localhost:3000`.
+
 ### Desktop application
 
 For Windows users, please install [Microsoft Edge WebView2][WebView2] before running the desktop app
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ docker-compose up
 ```
 Then, open a browser to `http://localhost:3000`.
 
+To stop and remove running containers, run `docker-compose down`.
+
 ### Desktop application
 
 For Windows users, please install [Microsoft Edge WebView2][WebView2] before running the desktop app
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,18 @@
+# dependencies
+/node_modules
+/.pnp
+.pnp.js
+
+# misc
+.DS_Store
+.netlify
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+/.eslintcache
+
+# vscode settings
+/.vscode
+/.husky/
+/src/redux/validation.json
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+FROM rust:1.51-buster
+
+RUN apt-get update && apt-get -y upgrade
+
+# Install nvm, node, npm, and yarn
+ARG NVM_DIR="/root/.nvm"
+ARG NODE_VERSION=14.6.0
+RUN curl https://raw.githubusercontent.com/creationix/nvm/v0.38.0/install.sh | bash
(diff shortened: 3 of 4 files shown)


thegrims:
Should add a .dockerignore file to reduce image size https://www.tutorialspoint.com/using-dockerignore-file
Could potentially add these files/directories in dockerignore
```
node_modules
.git
npm-debug.log
```

lekha:
I agree that the .git dir and npm-debug.log could be added to dockerignore, but I think node_modules should remain part of the image.

Pros:
* dependencies are the same across contributors
* dependencies are the same between local development and production (when production starts using this)

Cons:
* bigger images
* annoying image rebuild needed whenever dependencies are updated

For the cons, I think both points are addressable. Size-wise, the Rust libraries take the bulk (base `rust:1.51-buster` image is 1.25GB and final image is 1.84GB), so if this is a concern, it might be better to create a separate image for the desktop. On the annoying rebuilds, contributors can modify their docker-compose.yml to mount their local node_modules dir instead of mounting the one from the image, as I've done.

Do you still think node_modules should be in the dockerignore?

zakpatterson:
> I agree that the .git dir and npm-debug.log could be added to dockerignore, but I think node_modules should remain part of the image.
> 

This is what package-lock.json is for. It's supposed to guarantee the same set of node modules are pulled in.

README.md
```diff
@@ -58,6 +58,13 @@ This application can be either run as a web application or a standalone desktop
 
 To run, `npm start` or `yarn start`.
 
+If preferred, a Docker alternative is available:
+```
+docker-compose build
+docker-compose up
+```
+Then, open a browser to `http://localhost:3000`.
```
jswiss, line 66: Probably worth adding that users should run `docker-compose down` after exiting docker to shut down running containers.

jswiss:
If the purpose of this is to ensure that local development works the same for everyone, I'd argue that this PR ought to also introduce:
 - an `.nvmrc` file (thus invalidating that step in the Dockerfile);
 - a section in the README advising users how to work with [nvm](https://github.com/nvm-sh/nvm#nvmrc);
 - engines field in `package.json` specifying min/max Node and NPM versions.
 
 With these changes we'll be less likely to encounter issues with packages, peer dependencies, or tests not behaving as expected.

README.md
lekha: Done.

lekha:
@thegrims @zakpatterson I've added a .dockerignore that includes node_modules, but I realized we all had the wrong reasoning!

Ignoring the host node_modules dropped the images size from 1.85GB to 1.84GB, so this wasn't an issue about image size at all. And my previous comment about annoying rebuilds was true whether or not node_modules was in .dockerignore. The actual issue is that we don't want to mount our host node_modules, which might be on a different OS, to the container, which is running Debian buster.

Since size seemed like a concern, I went ahead and switched to a slim base, which dropped the image size 610MB.

lekha:
@jswiss I had initially used the .nvmrc file, but I still needed to add the node to my `$PATH`, which required hard-coding the Node version in the Dockerfile. This feels hacky, but I couldn't figure out the alternative. Can you elaborate how you expect it to be done?

zakpatterson (approved):

zakpatterson:
> Thinking about this a bit more, nvm and Docker seem to have conflicting ideologies. Docker focuses on environment isolation, while nvm assumes that a machine might have multiple versions of node. Since we only need _one_ version of node/npm in our Docker image, maybe I should install [directly from here](https://github.com/nodesource/distributions#installation-instructions) instead of using nvm at all. Thoughts?

@lekha So just have a single version of npm in Docker, and recommend nvm for people that don't want to use docker?

thegrims (approved):
Looks good! @lekha thanks for your contribution :)

PR #207, opened 2021-04-04
Fix material-ui label

zakpatterson:
If default required property is set on `<TextField>` the tooltip-style error is shown.

scrots_img_021


thegrims (approved):
LGTM

PR #208, opened 2021-04-04
Fix docs and package-lock for npm 6

zakpatterson:
Some of this might be handled by #206, but this fixes the immediate problem of npm 7 failing.

npm version: 6.14.12
node version: 15.12.0


thegrims (approved):
LGTM

PR #222, opened 2021-04-11
Handle undefined filing status, add task errors

zakpatterson:
Handle unselected filing status, also present task errors to user if they try to print a PDF with unselected filing status.

scrots_img_025


src/redux/reducer.ts
```diff
@@ -55,10 +55,18 @@ function formReducer (state: Information | undefined, action: Actions): Informat
       const newDependents = [...(newState.taxPayer?.dependents ?? [])]
       newDependents.splice(action.formData, 1)
 
+      const filingStatus = (() => {
```
thegrims, line 58: This function should probably be called something other than filingStatus to avoid confusion with the FilingStatus constant

src/pdfFiller/fillPdf.ts
```diff
@@ -78,6 +78,11 @@ export async function create1040 (): Promise<Uint8Array> {
 
   if (state.taxPayer !== undefined) {
     const f1040 = new F1040(state.taxPayer)
+
+    if (f1040.errors().length > 0) {
+      return await Promise.reject(f1040.errors())
```
thegrims, line 83: The await isn't required here

src/redux/reducer.ts
zakpatterson: See how it’s just a value though that gets used in the dictionary on 69

zakpatterson: It’s just a value though that gets used on L69?

thegrims: I guess it could just be personal preference. I'll resolve for now

src/pdfFiller/fillPdf.ts
thegrims: Nevermind!

thegrims (approved):

src/pdfFiller/fillPdf.ts
zakpatterson: Yea I didn’t either. Format on save just added it.

PR #223, opened 2021-04-12
Add f1040v voucher for taxes owed

zakpatterson:
If there is a tax amount owed on line 37 of the 1040, then the payment voucher 1040-v is attached to the front of the PDF. The voucher form has instructions about paying online, or cutting the voucher and including a check in the mail.


thegrims (approved):
LGTM

PR #229, opened 2021-04-12
Fix drawer to open/close on narrow screens, app bar, width consistency

### LINKED ISSUE #224: Fix main page container grid system
The form container grows horizontally if a contained component is too wide, for example if the `<LabeledInput>` has a long label. 

The custom css for the form component can be adjusted to fix this. Currently in `App.css`, the form width is set using `min-width`. Adding a `max-width` would prevent this from growing.

But we can do better. Also, on mobile the drawer navigation is not showing. So showing a menu button in mobile would be nice. At the same time we can address the form width containment.

Proposed goals are:

* Add app bar to top for narrow screens
* Add menu button to show / hide drawer for narrow screens
* Allow form to take up an appropriate amount of horizontal space depending on screen width.

https://material-ui.com/components/drawers/

zakpatterson:
This commit is a bit noisy due to indentation level changes. All the forms required a down tab.

Form container is now spaced to take up a comfortable amount of the screen width. This means we can just use global left align, and remove the need for many of the the wrapping `<Box>` components just to left align things.

Add app bar to top of form container.

Fixes #224


thegrims (commented):
A lot of this styling stuff is pretty subjective. With that being said I think the appbar should only show up on the mobile site. I also think the appbar on the mobile site looks cleaner without text content

thegrims:
image
section titles in the sidebar should be centered or aligned with the start of text in the section

zakpatterson:
@thegrims Do you like the constant width / wider form layout? That was the main motivation

thegrims:
@zakpatterson I like the constant width but I think I would prefer if the container were less wide and centered like a wider version of what exists currently. I think we could make this fullwidth version work but I think it might be nice to whiteboard / prototype some ideas for how the content could be layed out

zakpatterson:
Yeah the only reason to do it this way was to enforce a width that wasn't determined by the width of the children elements such as input labels. I will try it with just a wider gutter on wide screens.

thegrims:
Sounds good! I appreciate any improvements in the CSS layout structure, I know it's kind of set up in a weird way currently

thegrims:
image
The summary page is not aligned in width with the other pages

thegrims (approved):

thegrims:
Added a CSS change that permanently adds a scrollbar so that when page width increases because a dropdown is clicked, page elements no longer shift to the left.

https://stackoverflow.com/questions/1417934/how-to-prevent-scrollbar-from-repositioning-web-page

PR #234, opened 2021-04-13
Add display format for SSID

zakpatterson:
Choosing to use direct string replacement instead of reusing NumberFormat here, just for brevity

use here:

2021-04-13-221527_379x313_scrot

2021-04-13-221746_324x281_scrot


thegrims (approved):
LGTM

PR #235, opened 2021-04-13
Fix form warnings

zakpatterson:
There were numerous errors and warnings in the console we had been ignoring. This attempts to clear all those up. The three different errors/warnings types are separated by commit. See each commit for fixes for each type of warning.

### DIFF
--- a/src/components/input/LabeledDropdown.tsx
+++ b/src/components/input/LabeledDropdown.tsx
@@ -19,7 +19,6 @@ export function GenericLabeledDropdown<A> (props: LabeledDropdownProps<A>): Reac
               select
               helperText={error !== undefined ? 'Make a selection' : undefined}
               error={error !== undefined}
-              defaultValue=""
               SelectProps={{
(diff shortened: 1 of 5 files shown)


src/components/input/LabeledDropdown.tsx
```diff
@@ -19,7 +20,6 @@ export function GenericLabeledDropdown<A> (props: LabeledDropdownProps<A>): Reac
               select
               helperText={error !== undefined ? 'Make a selection' : undefined}
               error={error !== undefined}
-              defaultValue=""
```
thegrims, line 22: State is defaulting to Alabama with these labeledDropdown changes

thegrims: I think it should default to empty

zakpatterson: Yep, got that thanks

thegrims (approved):

PR #239, opened 2021-04-15
Dependabot updates

zakpatterson:


zakpatterson:
Ah tried opening this as an aggregation for merged dependabot PRs, but it's still causing netlify to build, so will close for now, but this would be a good way to track PRs merged into the staging branch.

PR #247, opened 2021-04-21
Implement Schedule E, real estate income and losses

zakpatterson:
This implements the first half of Schedule E, for real estate income and expenses. The second half of schedule E deals with K-1s received.

Schedule 1 is also added, required to get the net income from Schedule E forward to the 1040.

scrots_img_032


thegrims (commented):
Since the form can flex wider now using the grid, can the property financials be made flex to 2 columns to allow the page to be shorter?
image

README.md
```diff
@@ -15,8 +15,10 @@ Most income information from the following forms are supported:
 
 So far this project can attach the following schedules to form 1040:
 
+* Schedule 1
```
thegrims, line 18: Might want to add a qualifier that the only supported section of schedule 1 is real estate income

thegrims (commented):
image
when the foreign address box is checked, I think Province, Postal Code, Country should be shown instead of state, zip for property address

thegrims:
> image
> when the foreign address box is checked, I think Province, Postal Code, Country should be shown instead of state, zip for property address

Or since I see the se form mentions

> Physical address of each property (street, city, state, ZIP code)

if foreign addresses are not needed/desired, this checkbox should be removed from the rental income page

src/irsForms/ScheduleE.ts
```diff
@@ -0,0 +1,220 @@
...
+      p2 === undefined ? undefined : propTypeIndex[PropertyType[p2.propertyType]],
+      p0?.rentalDays, p0?.personalUseDays, p0?.qualifiedJointVenture,
+      p1?.rentalDays, p1?.personalUseDays, p1?.qualifiedJointVenture,
+      p2?.rentalDays, p2?.personalUseDays, p2?.qualifiedJointVenture,
+      [p0, p1, p2].find((p) => p?.propertyType === 'other')?.otherPropertyType ?? undefined,
+      ...(this.l3()),
+      ...(this.l4()),
+      ...(this.allExpenses().flat()),
+      ...(this.l19().flat()),
+      // ...(Array.from(Array(12)).map((_, i) => i))
```
thegrims, line 186: should this commented line be removed?

README.md
zakpatterson: :heavy_check_mark:

zakpatterson:
Good point on the foreign address, agree at this point we should just assume domestic addresses

thegrims:
When attempting to save new property, getting

> injectGlobalHook.js:526 Uncaught Error: Extension context invalidated.

and the property info is not successfully added to redux state

zakpatterson:
> When attempting to save new property, getting
> 
> > injectGlobalHook.js:526 Uncaught Error: Extension context invalidated.
> 
> and the property info is not successfully added to redux state

That's a weird looking error, haven't seen that before. Can you help reproduce? I'm not seeing it in the netlify. Maybe you could dump your localStorage to the console with `window.localStorage.getItem('persist:root')` and send me that?

zakpatterson:
> injectGlobalHook.js:526 Uncaught Error: Extension context invalidated.

It seems this message relates to a chrome browser extension?

thegrims:
image
Should 'Number of days in the year for personal use' be constrained to less than 366 - 'Number of days in the year used for rental'? Is the number of days in the year 366 to account for leap years?

zakpatterson:
> image
> Should 'Number of days in the year for personal use' be constrained to less than 366 - 'Number of days in the year used for rental'? Is the number of days in the year 366 to account for leap years?

2020 was a leap year, so max 366 is good. I updated it to change dependent on whether the year is a leap year.

I thought about adding another validator to prove rental days + personal use days <= 366. I can still do that.

zakpatterson:
> injectGlobalHook.js:526 Uncaught Error: Extension context invalidated.

Seems like this can be caused by an updated browser extension mismatching against some browser cache. I'm not seeing it on a clean chrome profile with the extension now. Maybe try reinstalling devtools if you're still seeing it?

https://github.com/facebook/react/issues/20811

thegrims:
image
Minor nitpick but parentheses indicate negative values on this sheet so for fields with parentheses the minus sign might be confusing

thegrims (commented):
image
Minor nitpick, parentheses indicate negative values on this sheet so for fields with parentheses the minus sign might be confusing

thegrims:
> 2020 was a leap year, so max 366 is good. I updated it to change dependent on whether the year is a leap year.
> 
> I thought about adding another validator to prove rental days + personal use days <= 366. I can still do that.

Probably good either way!

zakpatterson:
> Minor nitpick, parentheses indicate negative values on this sheet so for fields with parentheses the minus sign might be confusing

That's a good catch, fixed that row to display as positive, and another where zeros were being displayed.

thegrims (commented):
Found this https://turbotax.intuit.com/tax-tips/rental-property/real-estate-tax-and-rental-property/L3e09vT71

> 
> If I rent out my vacation home, can I still use it myself?
> Only for a very limited amount of time each year if you want the chance to fully deduct losses on your rental property. To be treated as a rental property for tax-loss purposes, your personal use of the place can't exceed 14 days or 10% of the days the unit is rented during the year, whichever is greater. While 10% may sound like a lot, it really isn't when you figure that a seasonal rental may only be in demand for two or three months each year.
> 
> If you violate the 14-day/10% rule, you can still deduct expenses associated with the rental, but only to the extent of your rental income.
> In other words, the property can't produce a net loss that will offset the income from other sources.

When I added >14 days of personal use per year and set usage personal usage equal to rental usage, rental income was incorrectly still able to be negative, functioning as a deduction
image

src/irsForms/ScheduleE.ts
```diff
@@ -117,10 +117,13 @@ export default class ScheduleE implements Form {
...
   ) as MatrixRow
 
-  l22 = (): MatrixRow => this.f8582.deductibleRealEstateLossAfterLimitation()
+  // Deductible real estate loss from 8582, as positive number
+  l22 = (): MatrixRow =>
+    this.f8582.deductibleRealEstateLossAfterLimitation()
+      .map((x) => x === undefined ? undefined : Math.abs(x)) as MatrixRow
 
   l24 = (): number => sumFields(this.l21().filter((x) => x !== undefined && x > 0))
   l25 = (): number => (
```
thegrims, line 129: I think l25 should also be absolute value
image

zakpatterson: This point about royalties has me confused and I think the issue is deeper than just filling some blanks here. What do you think about splitting off royalties as a separate issue and limiting this implementation to just real estate income? I'm researching more about this royalty income and it seems like it's not even pertinent to real estate, but more about intellectual property or something else.

thegrims: yeah that makes sense, royalties are not really logically grouped with rental income

zakpatterson: Ok took out royalties, and added absolute value on that always negative field, and parenthetical if negative on the line below.

thegrims: Did you see the other comment I left above WRT vacation housing vs rentals?

zakpatterson:
Implemented this test from the instructions, just on the schedule E directly. I think in the future we should stop the user from wasting time inputting expenses in the UI, but for now I think it's ok that the expenses are just left off the schedule e.

> Whether or not you can deduct expenses for the unit depends on whether or not you used the unit as a home in 2020. You used the unit as a home if your personal use of the unit was more than the greater of:
> 
>     14 days, or
> 
>     10% of the total days it was rented to others at a fair rental price.

src/irsForms/ScheduleE.ts
zakpatterson: Yep, just implemented that test, see below.

zakpatterson:
Added validation between the two number of days fields

scrots_img_034

thegrims (approved):
LGTM

PR #268, opened 2021-05-02
Fix earned income credit calculation function

zakpatterson:
Piecewise function evaluation was always selecting the upper (phase out) segment for any input.

Also add no income => no earned income credit test.

### DIFF
--- a/src/irsForms/ScheduleEIC.test.ts
+++ b/src/irsForms/ScheduleEIC.test.ts
@@ -12,7 +12,7 @@ jest.mock('redux-persist', () => {
   const real = jest.requireActual('redux-persist')
   return {
     ...real,
-    persistReducer: jest.fn().mockImplementation((config, reducers) => reducers)
+    persistReducer: jest.fn().mockImplementation((_, reducers) => reducers)
   }
 })
 
--- a/src/irsForms/ScheduleEIC.test.ts
+++ b/src/irsForms/ScheduleEIC.test.ts
@@ -16,12 +16,9 @@ jest.mock('redux-persist', () => {
   }
 })
 
-beforeAll(async () => {
-  jest.spyOn(console, 'log').mockImplementation(() => {})
+beforeAll(async () =>
   jest.spyOn(console, 'warn').mockImplementation(() => {})
-  jest.spyOn(console, 'info').mockImplementation(() => {})
-  jest.spyOn(console, 'debug').mockImplementation(() => {})
-})
+)
 
 describe('ScheduleEIC', () => {
   it('should disallow EIC for no income', () => {
--- a/src/irsForms/ScheduleEIC.test.ts
+++ b/src/irsForms/ScheduleEIC.test.ts
@@ -1,0 +1,52 @@
+import { waitFor } from '@testing-library/react'
+import { FilingStatus, PersonRole, TaxesState } from '../redux/data'
+import ScheduleEIC from './ScheduleEIC'
+import F1040 from './F1040'
+
+afterEach(async () => {
+  await waitFor(() => localStorage.clear())
+  jest.resetAllMocks()
+})
+
+jest.mock('redux-persist', () => {
+  const real = jest.requireActual('redux-persist')
+  return {
+    ...real,
+    persistReducer: jest.fn().mockImplementation((config, reducers) => reducers)
+  }
+})
+
+beforeAll(async () => {
+  jest.spyOn(console, 'log').mockImplementation(() => {})
+  jest.spyOn(console, 'warn').mockImplementation(() => {})
+  jest.spyOn(console, 'info').mockImplementation(() => {})
+  jest.spyOn(console, 'debug').mockImplementation(() => {})
+})
+
+describe('ScheduleEIC', () => {
+  it('should disallow EIC for no income', () => {
+    const model: TaxesState = {
+      information: {
+        f1099s: [],
+        w2s: [],
+        taxPayer: {
+          filingStatus: FilingStatus.MFJ,
+          dependents: [],
+          spouse: {
+            isTaxpayerDependent: false,
+            firstName: 'a',
+            lastName: 'b',
+            ssid: '123456789',
+            role: PersonRole.SPOUSE
+          }
+        }
+      }
+    }
+    const f1040 = new F1040(model.information.taxPayer)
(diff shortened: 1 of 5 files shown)


src/irsForms/ScheduleEIC.test.ts
```diff
@@ -0,0 +1,52 @@
...
+afterEach(async () => {
+  await waitFor(() => localStorage.clear())
+  jest.resetAllMocks()
+})
+
+jest.mock('redux-persist', () => {
+  const real = jest.requireActual('redux-persist')
+  return {
+    ...real,
+    persistReducer: jest.fn().mockImplementation((config, reducers) => reducers)
```
thegrims, line 15:
```suggestion
    persistReducer: jest.fn().mockImplementation((_, reducers) => reducers)
```

src/irsForms/ScheduleEIC.test.ts
```diff
@@ -0,0 +1,52 @@
+import { waitFor } from '@testing-library/react'
+import { FilingStatus, PersonRole, TaxesState } from '../redux/data'
+import ScheduleEIC from './ScheduleEIC'
+import F1040 from './F1040'
+
+afterEach(async () => {
+  await waitFor(() => localStorage.clear())
+  jest.resetAllMocks()
```
thegrims, line 8: If you reset mocks after each then I think your redux persist mock will go away after the first 'it' test. If another test is added that might be a problem

src/irsForms/ScheduleEIC.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    ...real,
+    persistReducer: jest.fn().mockImplementation((_, reducers) => reducers)
+  }
+})
+
+beforeAll(async () => {
+  jest.spyOn(console, 'log').mockImplementation(() => {})
+  jest.spyOn(console, 'warn').mockImplementation(() => {})
+  jest.spyOn(console, 'info').mockImplementation(() => {})
+  jest.spyOn(console, 'debug').mockImplementation(() => {})
```
thegrims, line 23: from looking at the scheduleEIC file, it looks like only console.warn needs to be mocked with a noop to silence output

src/irsForms/ScheduleEIC.test.ts
zakpatterson: Is that not desired? I don't think the second test should depend on the state resulting after the first test?

thegrims:
Thanks for adding a regression test!

src/irsForms/ScheduleEIC.test.ts
thegrims: You're right, I was thinking of `jest.restoreAllMocks`

src/irsForms/ScheduleEIC.test.ts
zakpatterson: Fixed this

src/irsForms/ScheduleEIC.test.ts
zakpatterson: I get your point now, yea per resetAllMocks doc:    
```
     * Resets the state of all mocks.
     * Equivalent to calling .mockReset() on every mocked function.
```
sounds ok

thegrims (changes requested):
Still getting slightly different results between preview environment and IRS EIC tool. This is with 1 dependent MFJ
image
vs 
image

https://apps.irs.gov/app/eitc

zakpatterson:
Ok, I will double check the formulas used. The earned income calculation is from here, page 10: https://www.irs.gov/pub/irs-drop/rp-19-44.pdf

zakpatterson:
I'm working through improving the code a little while trying to figure this out. Did you enter $10,000 for earned income into that calculator? Based on the procedure doc linked above,

scrots_img_029

MFJ+1 should allow EITC for earned income from `(0, 0)` to `(10540, 3584)`, so in that interval the EITC should be 34.003795%, or 10000*3584/10540 = $3,400.38

thegrims:
image
@zakpatterson here's a summary of what I entered to the IRS calculator

thegrims:
That calculation that you made checks out though, weird

thegrims:
Interestingly, If I make an income of $10,540 the IRS tool returns

>  you are eligible for the Earned Income Tax Credit estimated amount of: $3,579

Which is more than the maximum credit on that PDF you linked for 1 qualifying child

zakpatterson:
Ok, I got it. So the actual amount that should be paid is based on this table in publication 596:

scrots_img_030

Which says that any amount from 10,000 <= x < 10,050 gets $3409 as a credit. 

3409*10540/3584 = 10025. So this means that 10,000 income is rounded up to 10,025, but also 10,049 is rounded down to 10,025. 

So the optimal approach it would seem would be to round everyone's income mod 50 < 25 to the nearest 25, and anyone's income mod 50 > 25 calculate it based on the formula. Since the formula would show someone who earned $10,049 should get $3,417

zakpatterson:
> Interestingly, If I make an income of $10,540 the IRS tool returns
> 
> > you are eligible for the Earned Income Tax Credit estimated amount of: $3,579
> 
> Which is more than the maximum credit on that PDF you linked for 1 qualifying child

no, the max eitc for MFJ+1 per that doc above is $3584. The amount the calculator returned is based on the lookup table calculation, of income in the middle of a $50 window, 10525. 10525 * 3584/10540 = 3578.90

thegrims:
Gotcha, that makes sense! thanks for the explanation :)

zakpatterson:
Yeah what a pain to figure this out. I'll improve the code a bit and add a summary of this as a comment.

zakpatterson:
Ok, I added the rounding strategy I outlined above. I think it would be fine to just try to match the IRS online calculator, and the difference is never more than a few bucks. but I do think this rounding is technically correct. So as it stands incomes ending in 26-49 and 76-99 will not match the online calculator.

zakpatterson:
I found two more problems.

MFJ + 1 on the downward edge, $47000 earned income, shows there should be a credit but it's being disallowed.

Also for the rounding outlined above, it needs to be applied differently on the downward edge. I've changed my mind a bit, I'm just going to make this calculation match the IRS calculator as exactly as possible, then we can do the fancy rounding idea later.

thegrims (approved):
LGTM

PR #269, opened 2021-05-04
Fix regex error not accepting values < 10

zakpatterson:
nothing wrong with numbers between 0 and 10

scrots_img_031


thegrims (approved):
Good catch!

PR #282, opened 2021-05-09
Refactor forms to better separate concerns

zakpatterson:
I'm trying to better separate the PDF filling from the form assembly.

The functions in irsForms/Main complete all calculations and produce an array of forms, then pdfFiller/fillPdf takes that array of forms and pushes the data into the actual PDFDocuments.

Fixes an old factoring problem where there were two separate implementations of a function `getSchedules` that both tried to do the same thing.

This way we can create the complete data model for what will be inserted into the PDF, and probe it for correctness and for display.

Also changed return type for `create1040` so that the errors are returned explicitly.


thegrims (approved):
LGTM

PR #283, opened 2021-05-09
Fix Earned Income credit for real estate income

zakpatterson:
This uses pub 596 to bring net income forward from schedule E into the eligibility criteria for EIC. The EIC is precluded if there is more than $3650 of investment income. Rental properties must meet personal/rental use criteria to permit subtraction of expenses to reduce investment income.


thegrims (approved):
LGTM

PR #285, opened 2021-05-13
Merge filing status page into spouse and dependent page

thegrims:
fixes https://github.com/ustaxes/UsTaxes/issues/279
- adds filing status dropdown to spouse and dependent page
- removes filing status page, dropdown menu section, route in app
- removes react hook form reset call from dependent dropdown clear function because dependent cancel button was causing filing status to be cleared, and there was no difference in dependent dropdown clear functionality when it was removed
- adds strongLabel parameter to labeledDropdown


src/components/TaxPayer/SpouseAndDependent.tsx
```diff
@@ -211,15 +210,17 @@ const SpouseAndDependent = (): ReactElement => {
...
       { ({ onAdvance, navButtons }) =>
         <form onSubmit={handleSubmit(onSubmit(onAdvance))}>
-          <h2>Spouse Information</h2>
+          <h2>Family Information</h2>
+
+          <strong><p>Spouse Information</p></strong>
           <SpouseInfo />
 
-          <h2>Dependent Information</h2>
+          <strong><p>Dependent Information</p></strong>
```
zakpatterson, line 218: `<h2> or <h3>`?

zakpatterson (approved):

src/components/TaxPayer/SpouseAndDependent.tsx
thegrims: Can you elaborate on this comment? I'm not sure what it means

zakpatterson: Sorry, I just saw you removed an <h2> tag in favor of <strong><p>, seems like it should be a header tag, but it doesn't really matter.

thegrims: Oh yeah I removed <h2> in favor of <strong>  because I thought having one title <h2> and multiple <strong>  was a more clear layout. <h3> would definitely work as well. 

The reason I went with <strong> over <h3> is that I thought it might be cleaner to reuse the strongLabel logic rather than have a separate title

PR #301, opened 2021-05-18
Add property based testing, fix bug in segments

zakpatterson:
This adds [fast-check](https://github.com/dubzzz/fast-check) as a dependency and builds out one property based test.

Also, when fast-check detects an error, it attempts to simplify the inputs to the minimum complexity that triggers the same error. For example, in this PR, an error was spotted in the generic `segments` function.

This property tests that an array is divided into segments of equal length, with the excess at the last element. Below, I introduce a bug where the array is doubled in size if the input length is divisible by 9. A minimal breaking case is found in the property testing lengths:

```
    Property failed after 17 tests
    { seed: 1300705697, path: "16:60:0", endOnFailure: true }
    Counterexample: [[["[","\""," ","f","\"","=","x","\"","l"],2]]
    Shrunk 2 time(s)
    Got error: Error: expect(received).toEqual(expected) // deep equality

    Expected: 4
    Received: 9

    Stack trace: Error: expect(received).toEqual(expected) // deep equality

    Expected: 4
    Received: 9

      52 |           expect(s.length).toEqual(xs.length - expSegmentLength * (n - 1))
      53 |         } else {
    > 54 |           expect(s.length).toEqual(expSegmentLength)
         |                            ^
      55 |         }
```

Property based testing will help us ensure that more edge cases can be caught in our CI. Instead of writing individual unit tests, generating user data by hand, we can create arbitrary data models and test that 1040 outputs are sane and logical based on that data. 

For example, instead of creating a user model explicitly to test that someone with 3561 of investment income does not receive the EIC, we can generate random data models, and write `when investment income is > 3650, EIC is disallowed`, as one property of a valid 1040 given a model.


zakpatterson:
Added ability to generate random F1040s for testing.

thegrims (approved):
LGTM

PR #308, opened 2021-05-22
Set default file save name or directory

### LINKED ISSUE #307: Default file save name, extension, location
There should be a default file save name and location on desktop and web. For windows, a dialog pops including a blank file type. I assume if we just set a default file save name with `.pdf` as the extension then this will be fixed.

On the desktop version if we can select a default file directory, just use $HOME.

zakpatterson:
In browser mode, this saves the resulting PDF as `<last-name>-1040.pdf`

In desktop mode, this opens the save dialog window to the user's Documents directory, and sets the saving file type extension as pdf.

Fixes #307


zakpatterson:
> Just out of curiosity, why the difference between the desktop open-ended naming format vs templated web naming format?

I wasn’t able to figure out a default file name on the desktop side, just a default save directory. The opposite is true on the web side, a file name is set but the save directory is restricted. Not sure I understand your meaning.

thegrims:
I think that covers the question I had! Thanks :)

PR #354, opened 2021-06-13
Add Student Loan Interest Deduction

NathanielLovin:
This allows users to add Form 1098-E (which lists interest paid on student loans) and calculates the amount they can deduct from their income on Schedule 1 Line 20. Adds Form 1098-E data type to store those forms.


zakpatterson (approved):
I poked at this a lot, it looks great, nice work!

PR #365, opened 2021-06-23
Refactor to use FormContext

zakpatterson:
`react-hook-form`'s `useFormContext` from react-hook-form allows us to summon register and control at the location where they are actually needed, instead of drilling these properties all the way down our tree.

This context exists in `react-hook-form` v6, and this will greatly help the migration to `react-hook-form` 7, currently at version 7.9.0 (see #212). In react-hook-form 7, control is parameterized by the result type of the form.  Without this context, we would have to deal with that at various levels.

I've also simplified the number-format Pattern definitions so they no longer contain the control object, and simplified some code around pattern definitions.

### DIFF
--- a/src/components/Patterns.ts
+++ b/src/components/Patterns.ts
@@ -1,37 +1,28 @@
-import { Control } from 'react-hook-form'
 import { CURRENT_YEAR } from '../data/federal'
 import { daysInYear } from '../util'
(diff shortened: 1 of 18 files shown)


zakpatterson:
These changes were rolled into #375

PR #375, opened 2021-07-01
Upgrade react hook form

zakpatterson:
This includes #365 and supersedes #212 

This also eliminates the need for default values at every input. Instead, default values are managed by the controller, and in cases where the values need to be updated by force in the UI (for example to fill forms when an `edit` button is clicked) the Controller API is used (`reset`).


zakpatterson:
> filing status not getting persisted correctly, same with contact information, bank account and routing number

Sorry about that, just straight up missed applying the same refactoring for half the pages and the checkboxes. But it was pretty straight-forward. Time to get tests for more pages done.

thegrims:
> > filing status not getting persisted correctly, same with contact information, bank account and routing number
> 
> Sorry about that, just straight up missed applying the same refactoring for half the pages and the checkboxes. But it was pretty straight-forward. Time to get tests for more pages done.

No worries :)

thegrims (approved):
LGTM

PR #387, opened 2021-07-05
Fill previous values for FICA taxes in W2 edit control

aymarino:
In the 'Edit' control for an existing W2 entry, the fields for SS and Medicare tax withholding would be blank, even if the existing W2 data did have those fields set. It looks like the `defaultValue` property just wasn't set with the existing W2 data, like it is for the other fields.

### DIFF
--- a/src/components/income/W2JobInfo.tsx
+++ b/src/components/income/W2JobInfo.tsx
@@ -122,6 +122,7 @@ export default function W2JobInfo (): ReactElement {
         name="ssWithholding"
         patternConfig={Patterns.currency(control)}
         error={errors.ssWithholding}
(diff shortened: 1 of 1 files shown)


aymarino:
LOL, just saw #375:
> This also eliminates the need for default values at every input.

So this change might be superseded by that one

zakpatterson:
Nice catch, and yes we are changing how those values are passed soon based on updates to react-hook-form, but it just means moving the defaultValue settings up to the top of the render function.

Thanks for the contribution I will review this today!

zakpatterson (approved):

zakpatterson:
Thanks! I'll handle merging in #375 .

PR #413, opened 2021-07-09
Add additional Medicare tax (form 8959)

aymarino:
This adds checking and calculating additional Medicare tax, which kicks in at certain income thresholds and adds a flat 0.9% additional payroll tax. If income is above the threshold, this attaches form 8959 (Additional Medicare Tax) and Schedule 3 to the return, which reflect in the 1040 lines for total tax and total withholding.

In the simplest cases (single W2 filers), this doesn't actually change their taxes/refund owed, because the additional tax adds to their tax liability, but the extra payroll withholding is now also added to their taxes already paid, and the two should be the same. However, in a multiple W2 situation, the filer might owe additional tax as a result, because each W2 won't withhold enough at the higher rate. FICA has a lot of subtle rules.


zakpatterson:
>  FICA has a lot of subtle rules.

The main criticism I hear about this project is: "Oh the tax code is just too complex, with too many edge cases, and the big companies lobby government and make sure it stays that way. Capturing all that complexity in an open source application is just infeasible and we should probably leave it to large firms with lots of lawyers to do this for us."

Thank @aymarino for being an example of how open source is exactly the way to solve this problem! You know about some important edge cases, so you contribute the fix to solve those ones you know about. If enough people did this, we would reach total coverage for personal tax returns pretty quickly I think.

src/irsForms/F8959.ts
```diff
@@ -0,0 +1,138 @@
...
+    this.scheduleSE = scheduleSE
+  }
+
+  thresholdFromFilingStatus = (): number | undefined => {
+    const filingStatus = this.state.taxPayer.filingStatus
+    return filingStatus !== undefined ? fica.additionalMedicareTaxThreshold(filingStatus) : undefined
+  }
+
+  computeAdditionalMedicareTax = (compensation: number | undefined): number | undefined => {
+    return 0.009 * computeField(compensation)
```
thegrims, line 37: Thanks for your contribution! This should be rounded to an integer. The current result has a floating-point error

image

zakpatterson: Also I would add that .009 parameter deserves a place in the federal.ts file with the rest of our constants that will probably change year to year. 

`irsForms/util` has other functions for displaying numbers, also room for improvement in there, I think using those functions is a bit cumbersome to do math.

Good catch on the rounding issue, otherwise looks good to me.

aymarino: Good catch on the rounding issue! I also moved the tax rate to the data file for consistency, though this one actually doesn't really change year-to-year.

zakpatterson (approved):

thegrims (approved):

PR #417, opened 2021-07-11
Remove spouse W2s if filing status is MFS (fixes #416)

### LINKED ISSUE #416: Option should be removed to add spouse income when married filing separately is checked
currently when MFS is selected income can be added from a spouse, when that income should be added to the spouse's tax return, not the one being filed

zakpatterson:
Fixes #416 

* Splits W2 list based on employee
* Adds UI warning for MFS with spouse W2s
* Adds check in F1040 calculation to make sure we're only including relevant W2s based on filing
* Adds `react-if` as new dependency for simpler conditional rendering.

2021-07-11-121020_877x492_scrot


thegrims:
image
I think the TaxPayer Name's W2s title should only be displayed if that given payer has W2s inputted

thegrims:
image
Minor nitpick but could these W2s be changed to {spouse}'s W2s for clarity?

zakpatterson:
>  Minor nitpick but could these W2s be changed to {spouse}'s W2s for clarity?

I actually went back and forth for a minute between using the people's names and using "Your W2s" and "Spouse's W2s", then I just started getting annoyed that the IRS is still asking for "Your social security number" and "Spouse's social security number". Who is filling out the form? Just the primary and not the spouse? I think either would be, and in that case, who is the spouse? Aren't they both the spouse?

Anyway, no, I think we should use their names if we know them. I think changing it to "Primary taxpayer's W2s" and "Spouse's W2s" would be less confusing for us but more confusing for users.

zakpatterson:
Updated:

image

thegrims (approved):

PR #418, opened 2021-07-11
Add react-if and refactor complicated conditionals in jsx

zakpatterson:
This adds react-if and refactors a few confusing conditional places where the `<If>` component feels a little more natural. Also refactors the `<Summary>` component where it was not natural to use `If` but the render block was just too complex.

Because jsx `{ }`s can only contain expressions and not statements, to convert an if statement into an expression we had to use anonymous functions:

```tsx
<div>
  {(() => {
    if (condition) { return "info" }
    // no return === undefined, which is a valid ReactNode
  })()}
</div>
```

But instead of doing the above, we can just use `<If>`:

```tsx
<div>
  <If condition={condition}>
    info
  </If>
</div>
```

There are a couple downsides to this: 

* the block inside `<If></If>` will be evaluated regardless of the condition
* Union types are not resolved at compile time:

```tsx
type Result = undefined | string
const result: Result

if (result !== undefined) {
   result.length // ok
}

<If condition={result !== undefined}>
  {result.length} // compile time error
</If>
```


thegrims (approved):
LGTM

PR #419, opened 2021-07-12
Remove need to pass error to every field

zakpatterson:
The error value is already available from the form's context, and looking up the error only requires the name property, so we're able to remove a lot of duplicated work by just looking up the error where it is actually needed.

This PR should introduce no visible behavior changes. Adds `_.get` from `lodash` to get nested properties from `DeepMap`, for example if errors is `{'address': {'state': error_obj}}`, then `error_obj` can be accessed by `address.state`.


thegrims (changes requested):
See netlify comment

zakpatterson:
> Error messages not showing up for non number-format fields when fields are empty

I think I got it, this was an issue with errors populating in nested fields, e.g. `address.address -> {errors: {address: {address: ...}}`

thegrims (approved):

PR #420, opened 2021-07-12
Remove required=true on almost all form fields

zakpatterson:
A pattern object is passed which specifies the format the input can have. Passing this pattern object implies that the field is required, so specifying required=true on such fields is redundant.

On text inputs it makes required default true if `patternConfig` is defined.

Dropdowns are required default true.

### DIFF
--- a/src/components/income/F1099Info.tsx
+++ b/src/components/income/F1099Info.tsx
@@ -170,7 +170,8 @@ export default function F1099Info (): ReactElement {
 
   const intFields = (
     <LabeledInput
-      label="Box 1 - Interest Income" patternConfig={Patterns.currency}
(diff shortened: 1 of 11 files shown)


src/components/income/F1099Info.tsx
```diff
@@ -170,9 +170,7 @@ export default function F1099Info (): ReactElement {
 
   const intFields = (
     <LabeledInput
-      label="Box 1 - Interest Income"
-      required={true}
-      patternConfig={Patterns.currency}
+      label="Box 1 - Interest Income" patternConfig={Patterns.currency}
```
aymarino, line 173: Probably didn't mean to delete newline?

thegrims (approved):

PR #422, opened 2021-07-13
Add tests for FICA edge cases

aymarino:
It bothered me that the added logic for FICA edge cases didn't have any tests. I didn't add any because I didn't really understand the testing framework (the fact that the `arbitraries` were always random threw me), but I looked into fast-check and it's pretty slick. This adds some tests for those recently implemented FICA tax cases.


zakpatterson:
These look really good. Glad you like fast-check after trying it out. I've been struggling a bit to make clean tests using fast-check together with react-testing-library. It would be nice to be able to test the path from UI to model as well as from model to form.

thegrims (dismissed):

src/tests/fica.test.ts
```diff
@@ -0,0 +1,113 @@
...
+type Constructor<T> = new (...args: any[]) => T
+function hasAttachment<FormType> (attachments: Form[], formType: Constructor<FormType>): boolean {
+  return attachments.find((f) => { return f instanceof formType }) !== undefined
+}
+
+describe('fica', () => {
+  it('should give refund SS tax overpayment only in some conditions', () => {
+    fc.assert(
+      fc.property(arbitraries.f1040, ([f1040, forms]) => {
+        if (f1040.w2s.length <= 1) {
```
zakpatterson, line 31: Notice if you change the <= here, or the threshold, or any <= type operator, you won't get this test to fail.

the `expect` function returns a matcher object, but doesn't run the actual test. Note in the jest source: 

```
    interface Expect {
        /**
         * The `expect` function is used every time you want to test a value.
         * You will rarely call `expect` by itself.
         *
         * @param actual The value to apply matchers against.
         */
        <T = any>(actual: T): JestMatchers<T>;
```

the expect interface gives you methods like `.toEqual(x: any): void` that will throw if the assertion is invalid, so functions in here instead of just `expect(!hasSSRefund(f1040))` need to be `expect(hasSSRefund(f1040)).toEqual(false)`

zakpatterson (changes requested):

src/tests/fica.test.ts
thegrims: good catch!

aymarino: Ah yes, thanks for the catch. I'll find time to update it later today.

zakpatterson:
Good updates @aymarino. I'm making one point [over here on your branch](https://github.com/aymarino/UsTaxes/pull/2) about `.toBeTruthy()`, which I don't think we should use. If that makes sense feel free to merge there or I'll merge it here soon, then this should be good. Thanks!

zakpatterson (approved):

zakpatterson (approved):

zakpatterson:
So fast-check is finding a failure intermittently,  

It's possible to run exactly the failing case by adding the following object as a second object to `fc.assert`
```
{ seed: 1355933987, path: "36:0:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:1:2:2:3:2:2:2:4:2:2:2:2:2:2:2:2:3:3:2:2:2:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3:3", endOnFailure: true }```

The line 
```
              expect(hasSSRefund(f1040)).toEqual(false)
```
is failng

The counter example is:

```
[{
    "f1099s": [],
    "w2s": [{
        "occupation": "",
        "income": 0,
        "fedWithholding": 0,
        "employer": {
            "EIN": "051762887",
            "employerName": "",
            "address": {
                "address": "0 ",
                "aptNo": "0",
                "city": "",
                "state": "AL",
                "zip": "01416"
            }
        },
        "personRole": "PRIMARY",
        "ssWithholding": 0,
        "medicareWithholding": 0
    }, {
        "occupation": "",
        "income": 137701,
        "fedWithholding": 0,
        "employer": {
            "EIN": "098596990",
            "employerName": "",
            "address": {
                "address": "0 ",
                "aptNo": "0",
                "city": "",
                "state": "AL",
                "zip": "035091748"
            }
        },
        "personRole": "PRIMARY",
        "ssWithholding": 0,
        "medicareWithholding": 0
    }],
    "realEstate": [],
    "f1098es": [],
    "refund": {
        "routingNumber": "017054766",
        "accountNumber": "0227",
        "accountType": "checking"
    },
    "taxPayer": {
        "filingStatus": "S",
        "primaryPerson": {
            "firstName": "",
            "lastName": "",
            "ssid": "091126208",
            "role": "PRIMARY",
            "address": {
                "address": "0 ",
                "aptNo": "0",
                "city": "",
                "state": "AL",
                "zip": "092642926"
            },
            "isTaxpayerDependent": false
        },
        "spouse": {
            "firstName": "",
            "lastName": "",
            "ssid": "025511710",
            "role": "PRIMARY",
            "isTaxpayerDependent": false
        },
        "dependents": [],
        "contactEmail": "@.",
        "contactPhoneNumber": "0689910634"
    },
    "questions": {}
}]

```

aymarino:
Yeah I've been pretty impressed by fc, it found a similar gap on a boundary case in the other tests as well (which ended up being a place we didn't round properly on the actual form).

My local npm or node must be out of date maybe, based on the lint failing here. I'll circle back when I have a chance.

zakpatterson:
Hm the way GitHub chooses to render diffs for PRs in this situation is confusing. I'll review one more time soon but this looks pretty good.

zakpatterson (approved):

zakpatterson:
Just merged in master, and this looks great @aymarino.

thegrims (approved):

PR #429, opened 2021-07-18
Fix foreign country checkbox regression and make foreign country a dropdown input

### LINKED ISSUE #428: 'Do you have a foreign address?' checkbox not triggering input for foreign country and address

zakpatterson:
Fixes #428


thegrims (changes requested):
Country input should be a dropdown using data from the countries constant, rather than a freeform input

thegrims (approved):
LGTM, thanks!

PR #441, opened 2021-07-22
Add fields component to reduce repetition

zakpatterson:
Since we will have many many form fields, it might make more sense from here on to represent the form fields as a data structure instead of defining each individually in JSX. This may lead to benefits later, but for now it's at least more concise.

We might be able to expand on this a bit by including validation and model update logic in the field definitions, so that instead of each form requiring its own fully fleshed out functional react component, each field could exist independently. I think as our model data needs grow, the multi-page layout we have now might become a bit unwieldy and daunting for users. So this is a step in the direction of thinking about each piece of required data as a self contained entity rather than part of a form page.

For now, if the tradeoffs of this step seem beneficial, then we can explore ways to build on the idea later.

The main benefit for now is probably conciseness, which is really not much benefit at all if it reduces readability. So I'm proposing this idea just for comments!

### DIFF
--- a/src/components/Fields.tsx
+++ b/src/components/Fields.tsx
@@ -1,0 +1,62 @@
+import React, { ReactElement, Fragment, PropsWithChildren } from 'react'
+import { RegisterOptions } from 'react-hook-form'
+import { LabeledCheckbox, LabeledInput, USStateDropDown } from './input'
+import { CountryDropDown } from './input/LabeledDropdown'
+import { PatternConfig } from './Patterns'
+
+type FieldType = 'text' | 'country' | 'state' | 'checkbox'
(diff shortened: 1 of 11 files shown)


thegrims:
This would definitely cut down on repetitive <LabeledInput> declarations on pages. The concern I have is that it might make page design less flexible or that it might be difficult for developers onboarding to the project to pick up. Then again, I think you are right that the individual data / form elements should be logically seperate from the page form

bluSCALE4:
I've done a lot of work on this front and find it pretty enjoyable. I can help get this started or help in making the POC well rounded.

zakpatterson:
@bluSCALE4 You're welcome to contribute to this PR by opening a PR on the fork based on the branch it's coming from, assistance would be appreciated.

PR #459, opened 2021-07-30
Add prettier support

bluSCALE4:
Prettier saves time and sets a tone for code being committed. Some of the line breaks that get introduce might seem annoying at first but the level of consistency it brings will quickly win people over. We can do a mass application of prettier in a subsequent PR but this should get things going. I suggest adding the following to your VSCode config:

```
  "[typescriptreact]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode",
    "editor.formatOnSave": true,
  }
```

What is Prettier?

    An opinionated code formatter
    Supports many languages
    Integrates with most editors
    Has few options

Why?

    You press save and code is formatted
    No need to discuss style in code review
    Saves you time and energy

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,18 +1,16 @@
 module.exports = {
   settings: {
     react: {
-      "version": "latest",
-    },
+      version: 'latest'
+    }
   },
   env: {
     browser: true,
     es2021: true,
     node: true
   },
-  extends: [
-    'plugin:react/recommended',
-    'standard-with-typescript'
-  ],
+  extends: ['plugin:react/recommended', 'standard-with-typescript', 'prettier'],
+  parser: '@typescript-eslint/parser',
   parserOptions: {
     ecmaFeatures: {
       jsx: true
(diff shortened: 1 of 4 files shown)


zakpatterson:
We're already using eslint with the `standard` plugin. What does this add on top of that? 

https://github.com/ustaxes/UsTaxes/blob/master/.eslintrc.js#L14

zakpatterson:
Ok, I see this is in addition to eslint and the 'standard' plugin.

I think the best way to see its impact would be to just run it and incorporate its fixes in a second commit.

zakpatterson:
I was looking for the linter errors in the CI build, looks like the prettier plugin is missing from package.json:

https://github.com/ustaxes/UsTaxes/pull/459/checks?check_run_id=3198932953#step:5:13

bluSCALE4:
Yep, I'm fixing things and will push everything up once it's done.

bluSCALE4:
Having issues with @typescript-eslint/lines-between-class-members. Prettier is adding a line break to an inline function. I'm looking into something neutral to resolve this but it's tricky.

PR #460, opened 2021-07-30
Add Prettier support

bluSCALE4:
Prettier saves time and sets a tone for code being committed. Some of the line breaks that get introduce might seem annoying at first but the level of consistency it brings will quickly win people over. We can do a mass application of prettier in a subsequent PR but this should get things going. I suggest adding the following to your VSCode config:

```
  "[typescriptreact]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode",
    "editor.formatOnSave": true,
  }
```

What is Prettier?

    An opinionated code formatter
    Supports many languages
    Integrates with most editors
    Has few options

Why?

    You press save and code is formatted
    No need to discuss style in code review
    Saves you time and energy


bluSCALE4:
Sorry for the new PR, guess reverting commits may work a little differently on github.

zakpatterson (commented):
> Sorry for the new PR, guess reverting commits may work a little differently on github.

Not sure what you mean, but you probably just needed to force push to your branch to update the PR. It looks like you deleted the remote branch, and then pushed a new remote branch, which I think is also fine, but I think github closes your PR when you delete your remote branch, so maybe you could have just reopened the closed #459 PR.

No need to apologize, just trying to help you out for next time.

bluSCALE4:
The main takeaway is if a line is over 80 chars, it'll add breaks. It may add a lot but as I mentioned before, the consistency wins me over.

src/App.tsx
```diff
@@ -4,7 +4,7 @@ import './App.css'
 
 const App = (): ReactElement => (
   <div className="App">
-    <Main/>
+    <Main />
```
zakpatterson, line 7: This is good. I don't know why our linter isn't catching this.

src/components/FormContainer.tsx
```diff
@@ -15,17 +26,36 @@ const theme = createMuiTheme({
...
 
-const FormContainer = ({ onDone, onCancel, children }: PropsWithChildren<FormContainerProps>): ReactElement => (
+const FormContainer = ({
+  onDone,
+  onCancel,
+  children
+}: PropsWithChildren<FormContainerProps>): ReactElement => (
   <div>
     {children}
-    <Box display="flex" justifyContent="flex-start" paddingTop={2} paddingBottom={1}>
```
zakpatterson, line 21: I don't like this, attributes shouldn't be forced to a new line.

src/components/FormContainer.tsx
```diff
@@ -42,7 +72,14 @@ interface MutableListItemProps {
   icon?: ReactElement
 }
 
-export const MutableListItem = ({ icon, primary, secondary, remove, onEdit, editing = false }: MutableListItemProps): ReactElement => {
```
zakpatterson, line 45: Fine catch for long line.

src/components/FormContainer.tsx
```diff
@@ -83,7 +118,7 @@ export const MutableListItem = ({ icon, primary, secondary, remove, onEdit, edit
 }
 
 interface FormListContainerProps<A> {
-  onDone: (f: (() => void)) => () => void
```
zakpatterson, line 86: I like this.

src/components/FormContainer.tsx
```diff
@@ -131,19 +185,19 @@ const FormListContainer = <A extends object>(props: PropsWithChildren<FormListCo
...
-                primary={primary(item)}
-                secondary={secondary !== undefined ? secondary(item) : undefined}
-                onEdit={editAction(i)}
-                editing={editing === i}
-                remove={removeItem !== undefined ? () => removeItem(i) : undefined}
-                icon={icon !== undefined ? icon(item) : undefined}
-              />
-            )
-          }
+          {items.map((item, i) => (
```
zakpatterson, line 188: I'm not sure about this, I think I prefer the {} on new lines.

src/components/GettingStarted.tsx
```diff
@@ -2,75 +2,98 @@ import React, { ReactElement } from 'react'
 import { StartButtons, SingleButtons } from './pager'
 
 const repoUrl: string = 'https://github.com/ustaxes/UsTaxes'
-const codeOfConductUrl: string = 'https://github.com/ustaxes/UsTaxes/blob/master/docs/CODE_OF_CONDUCT.md'
```
zakpatterson, line 5: I think this is a good example of long strings causing long lines and that's ok.

src/components/Patterns.ts
```diff
@@ -56,20 +56,84 @@ const text = (regexp: RegExp, description: string): TextPattern => ({
...
+  ),
   name: text(/^[A-Za-z ]+$/i, 'Input should only include letters and spaces'),
-  zip: numeric(/[0-9]{5}([0-9]{4})?/, 'Input should be filled with 5 or 9 digits', undefined, undefined, '#####-####'),
-  ssn: numeric(/[0-9]{9}/, 'Input should be filled with 9 digits', undefined, undefined, '###-##-####'),
-  ein: numeric(/[0-9]{9}/, 'Input should be filled with 9 digits', undefined, undefined, '##-#######'),
-  currency: numeric(/[0-9]+(\.[0-9]{1,2})?/, 'Input should be a numeric value', undefined, undefined, undefined, '_', true, '$', 2),
-  bankAccount: numeric(/[0-9]{4,17}/, 'Input should be filled with 4-17 digits', undefined, undefined, '#################', ''),
-  bankRouting: numeric(/[0-9]{9}/, 'Input should be filled with 9 digits', undefined, undefined, '#########', '_'),
-  usPhoneNumber: numeric(/[2-9][0-9]{9}/, 'Input should be 10 digits, not starting with 0 or 1', undefined, undefined, '(###)-###-####'),
+  zip: numeric(
```
zakpatterson, line 83: I don't know, this is a weird set of definitions, and in this case, I think each on one line makes them more readable since they're all so similar.

src/components/Summary.tsx
```diff
@@ -153,13 +164,13 @@ const Summary = (): ReactElement => {
 
   return (
     <PagerContext.Consumer>
-      {({ navButtons, onAdvance }) =>
+      {({ navButtons, onAdvance }) => (
```
zakpatterson, line 167: I like using the parens, this is good.

thegrims:
@bluSCALE4 looks like eslint-plugin-prettier needs to be added to package.json

> 11:32:16 PM: ESLint couldn't find the plugin "eslint-plugin-prettier".
> 11:32:16 PM: (The package "eslint-plugin-prettier" was not found when loaded as a Node module from the directory "/opt/build/repo".)
> 11:32:16 PM: It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:
> 11:32:16 PM:     npm install eslint-plugin-prettier@latest --save-dev

src/components/TaxPayer/Address.tsx
```diff
@@ -10,13 +10,13 @@ interface AddressProps {
...
 
-export default function AddressFields (props: AddressProps): ReactElement {
+export default function AddressFields(props: AddressProps): ReactElement {
   const {
     checkboxText = 'Check if you have a foreign address',
     allowForeignCountry = true
   } = props
 
-  const { control } = useFormContext<{isForeignCountry: boolean }>()
+  const { control } = useFormContext<{ isForeignCountry: boolean }>()
```
zakpatterson, line 19: surprised our linter isn't catching this.

src/components/FormContainer.tsx
bluSCALE4: It only does it if it exceeds 80 characters.

zakpatterson:
> The main takeaway is if a line is over 80 chars, it'll add breaks. It may add a lot but as I mentioned before, the consistency wins me over.

But is prettier redundant to eslint+standard? `eslint --fix` already works, and vscode for example formats on save very fast with `eslint`. And we're already using [standard](https://standardjs.com/) which is meant to enforce a code style.

I like the idea of a line-limit in theory (thinking about 100 instead of 80), but there are situations where longer lines are unavoidable, so that means it's really up to developer discretion. I think our project has a ton of lazy long lines, but I'm not sure having a linter rule is the right way to fix them.  See [this issue](https://github.com/standard/standard/issues/28) for why standard.js decided not to include a line-length limit.

This PR though has exposed several inconsistencies in our code that I do think warrants revisiting code-style. I think we can switch to prettier if there's good reason to. I like the standardjs approach because the goal is to have zero style config in the project.

src/components/TaxPayer/Address.tsx
bluSCALE4: Yeah, it's little things like this that get me. As I mentioned, the line breaks are brutal at first but since it does its thing on save, you eventually don't even think about it which is the point. Without it, you leave it to the developer and you'll get mixed results and you'll be the linter in the PRs. For me it's not a tough call since I've used it for years, but it is an acquired taste.

zakpatterson:
Hey, @bluSCALE4, I've attempted to fix our eslint+standard config to match some of the changes that this config is picking up, and I can't make it work, so I think switching to prettier is a fine idea now.

I think you should be able to remove standard-ts and related dependencies as part of this PR. Can you give that a try when you have time? Also there's still a dependency missing in `package.json`. Check the log from the ESLint Action.

bluSCALE4:
Will look into the eslint log tonight.

thegrims (approved):

zakpatterson:
Because the diff is so large, I' have verified that purging ee5d47e and rerunning prettier produces a trivial diff with this branch, so that's good @bluSCALE4

One last issue though, 

After purging ee5d47e, I run `npm run lint` and there is no output, meaning success. Our CI lint action runs `npm run lint` (script from package.json) to make sure there are no linter errors. For a given commit, if prettier has work to do, then CI should fail. Also, `npm run lint:fix` does nothing, so that's not running prettier either.

Does a user have to have prettier installed globally (like `npm i -g prettier`) so they can just run `prettier --write .`? Is there a way to not require prettier to be installed globally so the prettier action can be done using `npm run prettier` or `npm run lint`?

zakpatterson:
Ok, I updated scripts in package.json

* `npm run prettier`: runs prettier formatter
* `npm run lint`: runs eslint and `prettier -c` to check if prettier needs to format. This is run by our linting CI action
* `npm run lint:fix`: runs eslint fix as well as `prettier --write`

zakpatterson (approved):

PR #463, opened 2021-07-30
Generate static pages for faster initial page load

zakpatterson:
This adds react-snap for an easy initial page load performance boost. Static rendered pages are served, with react back filling them with dynamic pages.

https://github.com/stereobooster/react-snap

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "ustaxes",
-  "version": "0.1.6",
+  "version": "0.1.8",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


thegrims:
image
when page is refreshed, navigation sidebar disappears

zakpatterson:
Interesting, I'm seeing that behavior on netlify only, and there's a CORS problem after initial load. 

But I'm also seeing a flash of the Getting Started page on refreshing the /info url, which is not expected. I'll investigate that a bit more.

thegrims:
image
Looks like form left padding was changed as well

zakpatterson:
Very close to working, for some reason on netlify it's now rendering a non-functioning, unstyled menu after refresh. Using `npm run serve ./build` locally (using `serve` from `npm`) seems to work pretty well though.

zakpatterson:
This is getting better, the sidebar doesn't disappear anymore on refresh. However there are still two issues:

* [ ] The routes are not all getting crawled, so subsequent routes are not getting generated. If the user refreshes an uncrawled page, then the page-not-found page will flash. This can be fixed by adding all routes to `reactSnap.include`
* [ ] The CSS is not static loaded properly. The page flashes with some styles added, but style changes when the dom is rehydrated.

@bluSCALE4 I think the second point is kind of in your domain, would you like to take a look?

PR #464, opened 2021-08-01
Initial implementation for form 1099-R

ctSkennerton:
This pull has an initial implementation for form 1099-R. Currently only the simplest case is implemented where the total value of 1099-R box 1 is added to 1040 line 4b.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ Most income and deduction information from the following forms are supported:
 - 1099-DIV
 - 1099-B
 - 1098-E
+- 1099-R: partial support for normal distributions from IRA accounts.
 
 So far, this project can attach the following schedules to form 1040:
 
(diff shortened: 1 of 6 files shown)


aymarino:
Recognizing that this is an initial simple implementation, it'd be nice to reflect what cases this does cover in the code, or better yet in the UI instructions too, and mark the rest as TODO.

It seems since this always takes the total distribution and enters it as taxable income in box 4b, it probably only will work for people taking normal distributions from IRAs (401k uses box 5a/5b). As a shortcut, I believe that case corresponds to distribution code 7 (in 1099-R box 7 -- https://www.irs.gov/pub/irs-pdf/i1099r.pdf page 16).

thegrims:
could the note about supported 1099 R income also be added to the readme

ctSkennerton:
Thank you for the tip about `npm run lint:fix` I was confused about those initial errors.

thegrims (approved):

PR #470, opened 2021-08-02
Add relative path

bluSCALE4:
Allows project to use src as relative path. For example, instead of './component/', you can use 'component/'


zakpatterson:
@bluSCALE4 

`git pull --rebase` followed by `git push --force` would have put your commit back on top of master and cleaned this up for you.

PR #471, opened 2021-08-02
Add baseUrl to tsconfig

bluSCALE4:
Add baseUrl to tsconfig to allow developers to access folders like components like this 'components/' instead of './components'.


thegrims (dismissed):

zakpatterson:
This should let us clean up some imports with `../../`, right? Can we do that in this PR too?

src/components/GettingStarted.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { ReactElement } from 'react'
-import { StartButtons, SingleButtons } from './pager'
```
zakpatterson, line 2: I think we should leave same package imports alone.

src/components/TaxPayer/PersonFields.tsx
```diff
@@ -7,10 +7,10 @@ import {
   ListItemSecondaryAction
 } from '@material-ui/core'
 import { useDispatch, useSelector } from 'react-redux'
-import { formatSSID, LabeledInput } from '../input'
-import { Patterns } from '../Patterns'
-import { Actions, removeDependent } from '../../redux/actions'
-import { TaxesState, Person } from '../../redux/data'
```
zakpatterson, line 13: This is good.

tsconfig.path.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "usTaxes/*": ["src/*"]
```
zakpatterson, line 5: is `us-taxes` or `ustaxes` more natural? This is kind of like an npm lib name, like `material-ui` or https://github.com/ustaxes/UsTaxes/blob/6ca9ad6a117b1c63ee8b29846e0dc21b5a735cb7/src/redux/store.ts#L6

zakpatterson:
@bluSCALE4 the paths setting seems to work for running the site, but it is failing in test. react-scripts seems to ignore it. I saw there's a work around involving https://github.com/gsoft-inc/craco. I haven't tried it yet.

zakpatterson:
This fails now due to `npm run build` failing:

```
11:56:37 AM: The following changes are being made to your tsconfig.json file:
11:56:37 AM:   - compilerOptions.paths must not be set (aliased imports are not supported)
11:56:38 AM: Creating an optimized production build...
11:56:46 AM: Failed to compile.
```

bluSCALE4:
> 
> 
> This fails now due to `npm run build` failing:
> 
> ```
> 11:56:37 AM: The following changes are being made to your tsconfig.json file:
> 11:56:37 AM:   - compilerOptions.paths must not be set (aliased imports are not supported)
> 11:56:38 AM: Creating an optimized production build...
> 11:56:46 AM: Failed to compile.
> ```

I did notice that. Typically paths are supported but I'm the build tools we're using don't.

PR #473, opened 2021-08-03
Move test files to src/test

eoncarlyle:
This moves two files that were outside of src/test into src/test


zakpatterson:
There's a util.test.ts in the root directory as well, let's do the same with that one

eoncarlyle:
@zakpatterson done

zakpatterson:
@eoncarlyle we recently added prettier as a linting step, so if your environment isn't set up to format with prettier on save, maybe try that. This just needs `npm run lint:fix`.

zakpatterson (approved):

PR #478, opened 2021-08-07
Add IL return, PDF -> code generator

zakpatterson:
Adds folders and stub files for every state
Adds a IL-1040 implementation

Adds UI for completing a state return if that state
has a return implementation available


zakpatterson:
Still a few bugs with fields from the IL return, and several fields are not yet completed.

zakpatterson:
There's currently an issue with the FilingStatus, Checking/Savings, and Nonresident/part-year resident checkboxes. These have radio functionality in the PDF, but they don't seem to be accessible as radiogroups in pdflib, and only the first checkbox is available in the array of fields from `getForm().getFields()`.

zakpatterson:
I added a typescript code generator that ingests a PDF, and outputs simple PDF code using the named form fields in the PDF. I successfully used it to build the IL1040 pdf code again. Needs more testing on more returns. To use, do:

`npm run formgen ./public/states/IL/IL-1040.pdf`, substituting your own PDF. It will output to console, which you can redirect to a file.

scripts/formgen.ts
```diff
@@ -0,0 +1,142 @@
...
+import { unzip } from '../src/util'
+
+const loadFile = async (path: string): Promise<PDFDocument> => {
+  const file = await readFile(path)
+  const bytearray = file.slice(0, file.byteLength)
+  return await PDFDocument.load(bytearray)
+}
+
+const normalizeName = (name: string): string =>
+  name.replaceAll(/[^a-zA-Z0-9]/g, '')
```
thegrims, line 13: Getting this error when running formGen
```
(node:15542) UnhandledPromiseRejectionWarning: TypeError: name.replaceAll is not a function
    at normalizeName (/mnt/c/Users/grims/github/UsTaxes/scripts/formgen.ts:13:8)
    at fieldFunction (/mnt/c/Users/grims/github/UsTaxes/scripts/formgen.ts:16:16)
    at Array.map (<anonymous>)
    at buildSource (/mnt/c/Users/grims/github/UsTaxes/scripts/formgen.ts:67:47)
    at /mnt/c/Users/grims/github/UsTaxes/scripts/formgen.ts:116:16
    at step (/mnt/c/Users/grims/github/UsTaxes/scripts/formgen.ts:52:23)
    at Object.next (/mnt/c/Users/grims/github/UsTaxes/scripts/formgen.ts:33:53
```
Might be because of lack of support? looks like a relatively new function (ES12)

zakpatterson: Interesting, yes it seems ts-node transpiles down and then your version of node doesn't have that function. `.replace` is fine here anyway.

thegrims (approved):
LGTM

PR #479, opened 2021-08-20
Expanded1099 r

ctSkennerton:
- New logic added in to separate 1099-R coming from 401K vs IRA which need to be reported differently
- Input now takes into account the gross distribution, taxable amount, and income tax withheld

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ Most income and deduction information from the following forms are supported:
 - 1099-DIV
 - 1099-B
 - 1098-E
+- 1099-R: partial support for normal distributions from IRA accounts.
 
 So far, this project can attach the following schedules to form 1040:
 
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Most income and deduction information from the following forms are supported:
 - 1099-DIV
 - 1099-B
 - 1098-E
-- 1099-R: partial support for normal distributions from IRA accounts.
+- 1099-R: support for normal distributions from IRA and pension accounts.
 
 So far, this project can attach the following schedules to form 1040:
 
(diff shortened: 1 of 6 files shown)


zakpatterson:
Awesome. There was a question about how this income affects the EIC eligibility. Have you checked on that while doing this amazing work?

ctSkennerton:
Sorry, I haven't looked at EIC.

ctSkennerton:
Though based on [this document](https://www.irs.gov/credits-deductions/individuals/earned-income-tax-credit/earned-income-and-earned-income-tax-credit-eitc-tables) it seems like it does not count as earned income

> Earned income does not include:

>Pay you got for work when you were an inmate in a penal institution
>Interest and dividends
>Pensions or annuities
>Social Security
>Unemployment benefits
>Alimony
>Child support

zakpatterson:
Ah thanks, I think that helps clarify. So if the 1099R is the persons only income then they would not be eligible, but if they had other earned income that qualified them, the 1099r would add to their AGI and reduce their credit.

zakpatterson:
It's behaving as expected with respect to EIC.

zakpatterson (approved):

thegrims:
@ctSkennerton Do the readme/start pages 1099-R description need to be updated to reflect these changes?

zakpatterson:
nice catch @thegrims, current README just references IRA accounts so additional doc descriptions there and the start page would  be nice.

ctSkennerton:
thanks for pointing it out. I've updated the readme and getting started guides

zakpatterson:
Looks good.

thegrims (approved):

PR #494, opened 2021-08-25
Dependabot updates

zakpatterson:
Need to check desktop build


zakpatterson:
Tried out desktop build, works normally.

thegrims:
The drop shadow for the form looks like it's getting cut off in the preview environment

zakpatterson:
> The drop shadow for the form looks like it's getting cut off in the preview environment

Weird but what in the diff could possibly cause that

zakpatterson:
It's being caused by this padding on the <main /> element, which comes from the `createMuiTheme`, conflicting with the box-shadow from our App.css file

So I think we need to bring the App.css stuff into the material-ui theme somehow.

image

thegrims:
I wonder if the padding changed for main, because the same issue is not showing up on the live site

thegrims:
alternatively, we could nix / modify the drop shadow

thegrims (approved):

thegrims:
LGTM

PR #506, opened 2021-08-26
Line 6: Social Security Benefits

ctSkennerton:
- Implement social security benefits worksheet that feeds into 1040 line 6
- Added in a new question that asks whether a person lived apart from
   their spouse, if they have social security income and they selected
   married filing separately.
- Added line 1040 6b to the list of fields that are used to calculate line 9
- Added SSA-1099 box 6 values into the formula that calculates 1040 line 25b


ctSkennerton:
I went and commented them out in the code since they are unused at the moment. Also added on SSA-1099 as one of the supported forms in the readme and getting started.

zakpatterson (approved):

thegrims (approved):

PR #508, opened 2021-08-27
Field focus

bluSCALE4:
Make form fields accessible by removing any issues that would prevent assistive technology user from completing any given task. Also shoe horned grid wrapper to form fields for responsive. This is great for mobile but might not look good once site functionality is increased.

Summary of work:

- **ButtonList padding** - Some of the minor changes include changing the spacing used for button lists from padding to margin. Margins can collapse if sibling elements also have margins. Paddings do not collapse so UI can look inconsistent at times.
- **ConditionallyWrap component** - Sometimes you want to wrap things. This is falling out of favor due to context api removing the need to wrap things but it is still useful for nested patterns like forms validations and layout components.
- **Increased Grid usage** - I'm a fan of using grids with material-ui. If we decide to move away from this pattern, it should be easy to revert.
- **Switched from NumberFormat prefix to MUI Input prefix** - NumberFormat didn't provide default spacing so I switched to MUI Inputs. The color schema matches labels and isn't auto selected so it looks more like a static prefix as opposed to auto inserted text
- **Remove custom labels** - Custom labels looked a little odd since the text fields were styled to include them. Opted to used MUI api for labels so labels are auto-connected to inputs. Also removed strong label and opted to use a component instead. It doesn't appear to be accessible as the strong semantics isn't being read out loud but I see it more as something to add contrast / emphasis than semantics.
- **Use TextField API to link error msgs** - Error messages now read out when fields are focused
- **Optimized Menu** - Footprint is now just <Menu />, fixed logic bug which used isMobile instead of drawer state.
- **Created Device Hook** - This will house functions like isPortrait and the like.

Form Preview:
Note the label and error are announced and all fields receive focus when enter (submit event) is hit. 
field_focus

Menu Preview:
resize

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+//.npmrc file
+engine-strict = true
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "build": "node ./setup.js && craco build",
     "test": "node ./setup.js && craco test",
     "eject": "craco eject",
+    "test:report": "node ./setup.js && craco test -- --coverage",
     "desktop-release": "node ./setup.js && tauri build",
     "desktop": "node ./notice.js && node ./setup.js && tauri dev",
     "tauri": "tauri",
@@ -99,6 +100,12 @@
   "jest": {
     "moduleNameMapper": {
       "^ustaxes/(.*)": "<rootDir>/src/$1"
-    }
+    },
+    "transformIgnorePatterns": [
+      "/node_modules/(?!@tauri-apps)"
+    ]
+  },
+  "engines": {
+    "yarn": "please-use-npm"
   }
 }
--- a/src/App.css
+++ b/src/App.css
@@ -1,24 +1,4 @@
-html {
-  overflow-y: scroll;
-  height: 100vh;
-}
-
 body {
   background-color: whitesmoke;
   height: 100vh;
-}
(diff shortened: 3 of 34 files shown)


thegrims (approved):

bluSCALE4:
Noticed defects with dynamic forms where selecting a dropdown option displays new fields. Currently, sibling Grids don't play nice. Wrapping Grids with a div should solve this but will look for a more elegant solution.

.npmrc
```diff
@@ -0,0 +1,2 @@
+//.npmrc file
+engine-strict = true
```
tmm1, line 2: What was the reasoning behind this? Was something not working with yarn?

bluSCALE4: Not sure but I figured if this was an npm project, why not restrict it to only use npm? Is this bad form? yarn and npm are quite different imo.

zakpatterson: From the [docs](https://docs.npmjs.com/cli/v8/using-npm/config#engine-strict):

> engine-strict
> 
>     Default: false
>     Type: Boolean
> 
> If set to true, then npm will stubbornly refuse to install (or even consider installing) any package that claims to not be compatible with the current Node.js version.
> 
> This can be overridden by setting the --force flag.
>

PR #514, opened 2021-08-30
Deduplicate list add/edit functions for editable list forms

### LINKED ISSUE #512: Issues loading saved data for properties and student loans
After adding an entry in Real Estate or Student Loan Interest, when the user clicks edit, the fields are not populated with stored values as expected.

zakpatterson:
FormListContainer is improved so it handles all the form resetting when an edit button is clicked. The user now only has to supply simpler onEdit, onSave, onDelete functions. Editing state is maintained by the generic component.

FormListContainer also does `handleSubmit`'s error handling and resetting of the edit form.

Additional testing is not provided in this PR yet, but the existing SpouseAndDependent does serve as a specific implementation of the FormListContainer so (copout) it does increase coverage since we're using this component more generically.

Fixes #512

### DIFF
--- a/src/components/FormContainer.tsx
+++ b/src/components/FormContainer.tsx
@@ -17,6 +17,11 @@ import {
 import { red } from '@material-ui/core/colors'
 import { Delete, Edit } from '@material-ui/icons'
 import { Else, If, Then } from 'react-if'
+import {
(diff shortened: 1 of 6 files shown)


thegrims (approved):
LGTM

PR #520, opened 2021-08-30
Add back typescript-eslint/recommended

zakpatterson:
Most notably this catches numerous unused imports and unused declared variables.

Many other rules catch questionable code, such as use of `{}`, object, any, String, Number as types. 

Removed `import React` per new recommended rules: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

### DIFF
--- a/src/components/ConditionallyWrap.tsx
+++ b/src/components/ConditionallyWrap.tsx
@@ -1,9 +1,7 @@
-import React, {
-  PropsWithChildren,
-  FunctionComponent,
-  ReactElement
-} from 'react'
+import { PropsWithChildren, FunctionComponent, ReactElement } from 'react'
 
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/no-explicit-any */
 const ConditionallyWrap = ({
   condition,
   wrapper,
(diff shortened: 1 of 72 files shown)


src/components/Summary.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { Fragment, ReactElement } from 'react'
+import { Fragment, ReactElement } from 'react'
```
bluSCALE4, line 1: Not sure if you want to just do away with all `Fragment`'s at this point.

zakpatterson: Agree, will add

bluSCALE4 (approved):

src/components/ConditionallyWrap.tsx
```diff
@@ -1,9 +1,7 @@
-import React, {
-  PropsWithChildren,
-  FunctionComponent,
-  ReactElement
-} from 'react'
+import { PropsWithChildren, FunctionComponent, ReactElement } from 'react'
 
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
zakpatterson, line 4: @bluSCALE4 I wasn't able to figure out a nice way to remove these disables, would you take a quick look

zakpatterson: One problem I see is that `PropsWIthChildren<{...children: ReactElement}>` is used, which duplicates the children key twice.

thegrims (approved):

PR #537, opened 2021-09-04
Add lodash and remove several util functions

zakpatterson:
* Using ES6: `anArrayOf(n, v)` => `Array(n).fill(v)` 
* Using lodash: `segments` => `_.chunk`
* Using lodash: `zip`, `zip3` => `_.zipWith`
* Using lodash: `unzip` => `_.unzip`


bluSCALE4:
We should utilize tree shaking and import modules individually. https://lodash.com/per-method-packages

zakpatterson:
> We should utilize tree shaking and import modules individually. https://lodash.com/per-method-packages

I think this link shows you don't need to do it this way anymore? I think our builder does tree shaking for us?

bluSCALE4:
> 
> 
> > We should utilize tree shaking and import modules individually. https://lodash.com/per-method-packages
> 
> I think this link shows you don't need to do it this way anymore? I think our builder does tree shaking for us?

Fair enough. I'll add https://www.azavea.com/blog/2019/03/07/lessons-on-tree-shaking-lodash/ to make sure things are lean.

zakpatterson:
Ah, is your idea that we should `'import {function names} from 'lodash'` instead of just importing the whole thing? I thought you were talking about the napm micro packages. 

I'm not sure that makes a difference anymore. I also don't think we would prioritize bundle size above readability. But I would be interested to know if it still makes a difference. I'm hoping the typescript compiler and babel transforms have progressed far enough. Anyway I wouldn't worry about this at this point I think.

zakpatterson:
If there's another lib that would offer these functions, I'd like to use that. I think lodash is a bit outdated at this point.

zakpatterson:
fp-ts looks really good

https://gcanti.github.io/fp-ts/modules/Array.ts.html

zakpatterson:
I found this article that explains how depending on use of lodash, you can break treeshaking completely.

https://medium.com/bootstart/why-using-chain-is-a-mistake-9bc1f80d51ba

Basically lodash has a `chain` function, which necessarily exposes the entire library. So even if you just write `_.chain(xs).map(...).value()`, the chain call brings in every available function. So that's pretty convincing, plus I think the style suggested in that article is better anyway.

thegrims:
> I found this article that explains how depending on use of lodash, you can break treeshaking completely.
> 
> https://medium.com/bootstart/why-using-chain-is-a-mistake-9bc1f80d51ba
> 
> Basically lodash has a `chain` function, which necessarily exposes the entire library. So even if you just write `_.chain(xs).map(...).value()`, the chain call brings in every available function. So that's pretty convincing, plus I think the style suggested in that article is better anyway.

Are you going to modify the code to fit the style outlined in the article? Or will that go in a follow up PR?

zakpatterson:
> Are you going to modify the code to fit the style outlined in the article? Or will that go in a follow up PR?

I think at this point no. The flow functions suggested there don't really make any sense for typescript. If you write `flow(f, g, h)(data)` then the typechecker has no idea what f and g and h are, since data comes last. So not a fan of that now. 

I think the current way is good for now, but someone will probably be offended and motivated to improve it down the line. Not a bad thing.

zakpatterson:
> I found this article that explains how depending on use of lodash, you can break treeshaking completely.
> 
> https://medium.com/bootstart/why-using-chain-is-a-mistake-9bc1f80d51ba
> 
> Basically lodash has a `chain` function, which necessarily exposes the entire library. So even if you just write `_.chain(xs).map(...).value()`, the chain call brings in every available function. So that's pretty convincing, plus I think the style suggested in that article is better anyway.

Actually now I don't think so at all, consider the "bad" example:
```
import _ from "lodash"
_.chain([1, 2, 3])
  .map(x => [x, x*2])
  .flatten()
  .sort()
  .value()
```

Now the 'good' example
```
import map from "lodash/fp/map"
import flatten from "lodash/fp/flatten"
import sortBy from "lodash/fp/sortBy"
import flow from "lodash/fp/flow"
flow(
  map(x => [x, x*2]),
  flatten,
  sortBy(x => x) 
)([1,2,3])
```

Except the good example can't typecheck at all. `flow` actually takes `any`s because at that point there's nothing to indicate what flow will operate on. map, flatten, and sortBy have no type information attached to them. Flow is also just the same as reduce, like:

thegrims (approved):

Issue #481, opened 2021-08-23
Support apostrophes in names


eoncarlyle:
The application should be able to support apostrophes in names, which can be common for Irish-American surnames. Changing the form validation to address this should be straightforward

![image](https://user-images.githubusercontent.com/12573703/130379769-db973d61-ac7c-41be-9625-26b18f961cd8.png)

zakpatterson:
It would be nice to find a document supporting what exact characters are allowed. For example what characters, if any, not used in the English language are allowed?  I would think hyphens should also be allowed.

thegrims:
> The application should be able to support apostrophes in names, which can be common for Irish-American surnames. Changing the form validation to address this should be straightforward
> 
> ![image](https://user-images.githubusercontent.com/12573703/130379769-db973d61-ac7c-41be-9625-26b18f961cd8.png)

Good catch!

eoncarlyle:
@zakpatterson page 103 of the [California Birth Registration Handbook](http://www.avss.ucsb.edu/MANUALS/Handbook.pdf) provides hyphens, apostrophes, periods, and commas as acceptable punctuation - what are our thoughts on using these rules?

thegrims:
I could also just remove the validation regex so that the field accepts any text and leave it up to the user to enter their name correctly

zakpatterson:
> I could also just remove the validation regex so that the field accepts any text and leave it up to the user to enter their name correctly

Not having found any restriction in the IRS instructions, I think this is the way.

PR #543, opened 2021-09-07
removing name regex from first and last name fields

thegrims:
Addresses #481


zakpatterson:
I saw your failed tests so I tried to take care of that. Instead of specifying a `patternConfig`, just `required={true}` is all that's needed, then we can fix up the tests to match the new requirements.

thegrims:
@zakpatterson Thanks!

PR #548, opened 2021-09-08
Add react helmet and titles

bluSCALE4:
Update pages so titles are updated on transitions.


thegrims (approved):

thegrims (changes requested):
Seeing this error in the unit test
`Invariant failed: You should not use <Prompt> outside a <Router>`

src/components/deductions/F1098eInfo.tsx
```diff
@@ -87,12 +88,15 @@ export default function F1098eInfo(): ReactElement {
   )
 
   return (
-    <form tabIndex={-1} onSubmit={onAdvance}>
-      <FormProvider {...methods}>
+    <FormProvider {...methods}>
+      <form tabIndex={-1} onSubmit={handleSubmit(onAdvance)}>
+        <Helmet>
+          <title>1098-E Information | Deductions | UsTaxes.org</title>
```
zakpatterson, line 94: Would it be better for all these titles to be in one place in the [Main component](https://github.com/ustaxes/UsTaxes/blob/master/src/components/Main.tsx#L122)

thegrims (approved):

PR #559, opened 2021-09-12
Add in W2 box 5.

ctSkennerton:
- Box 5 is the medicare income, which may be different from box 1.
- Change F8958 line 1 to use W2 box 5 as instructed.
- Fixes bug where the additional medicare tax calculated in F8959 is wrong if W2 box 5 does not equal W2 box 1.


zakpatterson:
There is a failing test, and fast-check is printing out a ton of compiled javascript so it looks confusing. I put up https://github.com/ustaxes/UsTaxes/pull/561 to fix this.

ctSkennerton:
thanks. I think it's because the test is trying to calculate the expected value based on the income, rather than the medicare income. But, as you say, the output is extremely verbose to really tell anything.

ctSkennerton:
yep, that seemed to do the trick. There were a couple of places where threshold was still being calculated based on the wages rather than medicare wages.

thegrims (approved):

PR #560, opened 2021-09-12
State income withholding

zakpatterson:
* Adds UI for W2 boxes b, 15, 16, 17.
* Fixes error in PDF -> TS script
* Adds withholding to IL1040
* Adds IL1040 Schedule WIT (only W2 withholding supported)
* Supports adding multiple WITs if there are more than 5 withholding forms for a single ssid


src/components/income/W2JobInfo.tsx
```diff
@@ -47,18 +67,31 @@ const toIncomeW2 = (formData: IncomeW2UserInput): IncomeW2 => ({
...
   fedWithholding: data.fedWithholding.toString(),
   ssWithholding: data.ssWithholding.toString(),
-  medicareWithholding: data.medicareWithholding.toString()
+  medicareWithholding: data.medicareWithholding.toString(),
+  state: data.state,
+  stateWages: data.stateWages?.toString() ?? '',
+  stateWithholding: data.stateWithholding?.toString() ?? ''
 })
 
+const boxLabel = (box: string, description: string): ReactElement => (
```
thegrims, line 89: since other pages use a box with the 'strong' label, I think this should be moved to be in the input directory

src/components/income/W2JobInfo.tsx
```diff
@@ -173,12 +200,23 @@ export default function W2JobInfo(): ReactElement {
...
-          label={
-            <>
-              <strong>Box 6</strong> - Medicare tax withheld
-            </>
-          }
+          label={boxLabel('6', 'Medicare tax withheld')}
+          patternConfig={Patterns.currency}
+          sizes={{ xs: 12, lg: 6 }}
+        />
+        <USStateDropDown name="state" label={<strong>Box 15 - State</strong>} />
```
thegrims, line 207: only box and the number should be bolded in this label

thegrims (approved):

PR #565, opened 2021-09-17
Add page to save and load user data to json file

zakpatterson:
Let the user save their data to a file, and read it back out again from a file.

As with the localstorage object, no real sanity checks are applied yet.

### DIFF
--- a/src/redux/External.tsx
+++ b/src/redux/External.tsx
@@ -1,0 +1,83 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { Button, ButtonProps } from '@material-ui/core'
+import { ChangeEvent, PropsWithChildren, ReactElement } from 'react'
+import { useDispatch, useStore } from 'react-redux'
+import { Information, TaxesState } from './data'
+import { setEntireState } from './actions'
+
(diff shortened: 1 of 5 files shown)


thegrims:
image
can the settings button color be changed to be the same as the other icons?

zakpatterson:
> image
> 
> can the settings button color be changed to be the same as the other icons?

Ah sure I didn't check it in dark mode.

zakpatterson:
image

src/redux/External.tsx
```diff
@@ -0,0 +1,83 @@
...
+const Exfiltrate = (props: PropsWithChildren<ButtonProps>): ReactElement => {
+  const store = useStore<TaxesState>()
+
+  const { children, ...rest } = props
+
+  const info = store.getState().information
+
+  const onClick = () =>
+    download(
+      `${info.taxPayer.primaryPerson?.lastName ?? 'return_data'}.json`,
```
thegrims, line 72: can the file name contain something like ustaxes_return_data regardless of the last name? I think that would make it more clear for users what the file is used for
```suggestion
      `ustaxes_return_data.json`,
```

PR #566, opened 2021-09-17
Support multiple tax years

zakpatterson:
By overriding useSelector and useDispatch, and making small change to the action types and reducers, we can avoid modifying component code.

Our `TaxesState` global state object now will contain a different copy of the entire data model for each tax year. So instead of 

{information: Information}, it's 

`{ Y2019: Information, Y2020: Information, Y2021: Information, activeYear: TaxYear}`

Where updating `activeYear` changes the effect of all actions.

Hopefully, no modification to UI pages will be needed to support this, except for when we realize different questions must be asked for each year, or different data elements are required for different years.

Goals for this PR:

* [x] Update data model to support other tax years
* [x] Change action types to support tax years
* [ ] Add UI to change current viewing tax year
* [ ] Add UI to propagate prior year data into subsequent year
* [ ] Check tax form definitions for 2019


zakpatterson:
I don't know which is preferable yet. It would be nice if there were just one action needed to change the year. On the other hand if the user has unsaved data it will be wiped out by the change and there's no confirmation nag, so it's kind of a jarring update. I used the stateful button to try to show that some change was applied by the selection.

It's still not good enough because there's too much separation between it and what it's modifying in the UI.  I think there should be some header bar above the data display that shows something like 

"You're viewing tax data for 2020"

And the 2020 would either be an select component that changes instantly or text that requires a couple clicks to make that change. I think the UI could easily be made too confusing so the user mistakenly adds data to the wrong year.

Issue #465, opened 2021-08-02
Require signed commits


zakpatterson:
Thinking about code security, from some point going forward we should require signed commits. 

https://withblue.ink/2020/05/17/how-and-why-to-sign-git-commits.html

This will be a little annoying for new contributors' PRs, but when a user sends a new PR, if commits are unverified, we would ask them to follow [this tutorial](https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification), apply `git config --global commit.gpgsign true`, and then `git rebase master`, followed by `git commit --amend` to redo each commit on the branch, followed by force pushing.

We can add this to CONTRIBUTING as well as to a PR template.

Other projects such as Tauri require verified commits.

thegrims:
Thanks for the link! Good thing to be aware of

PR #586, opened 2021-09-26
Estimated tax payments

ctSkennerton:
Here is my initial pass at having a page for payments and implementing estimated taxes for form 1040 line 26.


zakpatterson:
Hey I saw this oddly was passing the linter action but failing when trying to deploy, due to two linter warnings. So I fixed the `npm run lint` command to disallow all warnings and cleaned those two up.

zakpatterson:
I'm going to start gently nudging people to set up signed commits. I think it's a good step to take for code security:

https://github.com/ustaxes/UsTaxes/issues/465

No problem on this PR, but at some point in the near future we'll start requiring it.

ctSkennerton:
Thank you @zakpatterson! I added in a gig key so my future commits should be properly verified.

thegrims:
I think it might be good to start off estimated taxes with a question prompt of 

> did you already make payments towards your 2020 taxes this year or last year

or similar, to avoid confusion of what this field is asking for

ctSkennerton:
I've updated the code to fix the issue with saving and also added in a label field in case a user has multiple payments and wants to provide a note for what each one is. I think the CI is running into the same issue as documented in #588, which is causing it to fail.

zakpatterson:
> I've updated the code to fix the issue with saving and also added in a label field in case a user has multiple payments and wants to provide a note for what each one is. I think the CI is running into the same issue as documented in #588, which is causing it to fail.

Agree. Reran twice and it passed.

thegrims (approved):

zakpatterson:
Looks good, not going to hold this up for testing given #588.

PR #698, opened 2021-10-30
Add state reset and error for filing status

### LINKED ISSUE #697: Error / state handling incorrect for filing status
Steps to reproduce:

1. Save filing status as single
2. Add spouse  [Observe that filing status is blanked out (no selection)]
3. Save page  [Observe page transition, and filing status single is set as a redux action]

Single is not a valid filing status if a spouse is present, so an error should be displayed in the UI, and the action should not be sent.

It would be nice if the action creator could block such an action from being created when a spouse is present.

zakpatterson:
If user has single saved and then adds a spouse, we show an
error and remind them to select a new status.

image

If a user has MFJ selected and deletes a spouse, we show an
error and remind them to select a new status.

image

Form state in these cases is now reset to empty, and this way
react-hook-form properly shows error message and prevents
submitting data.

Fixes #697 

- Bugfix

### DIFF
--- a/src/components/TaxPayer/SpouseAndDependent.tsx
+++ b/src/components/TaxPayer/SpouseAndDependent.tsx
@@ -1,4 +1,4 @@
-import { ReactElement } from 'react'
+import { ReactElement, useEffect, useState } from 'react'
 
 import { useForm, FormProvider } from 'react-hook-form'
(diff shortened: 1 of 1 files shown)


thegrims (approved):
LGTM! Thanks

PR #699, opened 2021-10-31
Improve UX when adding/editing records

### LINKED ISSUE #549: Improve UX for adding records
Right now, when a user is adding or modifying a record, is the user hits enter or clicks next / previous, data will be lost. 

Scenario 1
Current Behavior:
User clicks add / edit
User fill out information
User presses enter
Data is lost

Expected Behavior:
User clicks add / edit
User fill out information
User presses enter
Record is saved

Scenario 2
Current Behavior:
User clicks add / edit
User fill out information
User clicks prev or next
Data is lost

Expected Behavior:
User clicks add / edit
User fill out information
User clicks prev or next
User is prompted that data will be lost if record is in dirty state.

dbelokon:
- Other, please describe: UX improvement
- Minor test adjustments

Fixes #549.

I also had to modify certain tests because some of them would fail due to the usage of `<Prompt>` outside of `<Router>` component. I went ahead and modified those tests so that the test subjects are wrapped by a `<BrowserRouter>`. The error shown was very similar to [this error discussed in a StackOverflow discussion](https://stackoverflow.com/questions/62710458/invariant-failed-you-should-not-use-switch-outside-a-router).

### DIFF
--- a/src/components/FormContainer.tsx
+++ b/src/components/FormContainer.tsx
@@ -1,4 +1,4 @@
-import { PropsWithChildren, ReactElement, useState } from 'react'
+import { PropsWithChildren, ReactElement, useState, KeyboardEvent } from 'react'
 import {
   createStyles,
   makeStyles,
@@ -29,8 +29,17 @@ const FormContainer = ({
   children
 }: PropsWithChildren<FormContainerProps>): ReactElement => {
   const prefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)')
+
+  function handleEnterKey(event: KeyboardEvent<HTMLDivElement>) {
+    const target = event.target as HTMLInputElement
+    if (target != null && event.key === 'Enter') {
+      event.preventDefault()
+      onDone()
+    }
+  }
+
   return (
-    <div>
+    <div onKeyDown={handleEnterKey}>
       {children}
       <Box
         display="flex"
--- a/src/components/FormContainer.tsx
+++ b/src/components/FormContainer.tsx
@@ -1,11 +1,4 @@
-import {
-  PropsWithChildren,
-  ReactElement,
-  useState,
-  KeyboardEvent,
-  useCallback,
-  useEffect
-} from 'react'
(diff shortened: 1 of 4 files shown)


src/components/FormContainer.tsx
```diff
@@ -263,11 +296,19 @@ const FormListContainer = <A,>(
 
   return (
     <>
+      <Prompt
+        when={formState === FormState.Dirty}
```
zakpatterson, line 300: Did you run across `react-hook-form`'s `isDirty`? 

It should be in scope already, in the above line:

`const { reset, handleSubmit, formState: {isDirty} } = useFormContext()`

Maybe you already saw that and there's a reason it doesn't fit this situation?

thegrims:
Created a new issue [If foreign address is checked, it incorrectly remains checked even if it is unchecked #700](https://github.com/ustaxes/UsTaxes/issues/700) in GitHub

src/components/FormContainer.tsx
```diff
@@ -263,11 +296,19 @@ const FormListContainer = <A,>(
...
+        message="Warning: you have unsaved changes. Are you sure you want to leave?"
```
thegrims, line 301: I'm not seeing this message show up in the preview environment, is there a section/page I should visit?

zakpatterson: It looks like it just shows up on pages with the internal item forms, like 1099s, W-2s, spouseanddependent, if the user navigates away using the sidebar while unsaved data is present in the form.

image

thegrims: Gotcha

src/components/FormContainer.tsx
zakpatterson: I went ahead and switched the implementation to use the existing `FormContext`'s `isDirty`, and it seems to work the same.

zakpatterson (approved):

thegrims (approved):
LGTM

thegrims:
@dbelokon Thanks for your contribution!

dbelokon:
You are welcome!! 😁

PR #717, opened 2021-11-10
Fix floating point errors in PDFs

zakpatterson:
- Bugfix
- Refactor

Rounding formatting errors are present in PDF fields with numeric values. `.30000000004` style errors as well as string formatting such as `123.6` are also present.

This also refactors things somewhat so we can provide a different downloader in testing. This will also be useful in #686 where pdfs are downloaded from ustaxes.org instead of from the testing machine file system.

See 611de2a for output from failing test.


thegrims (approved):
LGTM!

PR #826, opened 2022-01-02
Fix not required values 701

### LINKED ISSUE #701: Not required values in form being pushed to state
1. Add a W2, with nothing entered for state wages
2. click edit 
3. click save
4. observe a "numeric value" error under the state wages field.

The data is being pushed to the redux store as NaN

image

There is a schema error about this, but the schema error does not prevent the data from being stored.
image

zakpatterson:
Fixes #701

I tried for far too long to get a failing test that would detect NaN showing up in the dispatched action or in the state, but failed. I'll have to revisit that.

- Bugfix

### DIFF
--- a/src/components/income/W2JobInfo.tsx
+++ b/src/components/income/W2JobInfo.tsx
@@ -28,7 +28,7 @@ import { Grid, Box } from '@material-ui/core'
 import { Work } from '@material-ui/icons'
 import { addW2, editW2, removeW2 } from 'ustaxes/redux/actions'
 import { Alert } from '@material-ui/lab'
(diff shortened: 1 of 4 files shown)


thegrims:
image
stateWages are still allowed to be undefined

zakpatterson:
> stateWages are still allowed to be undefined

Initially I thought this was ok, because W-2s received for non-filing state residents might be blank in this section? But someone that moved during the year might have a certain amount attributed to each state. So really the state wages /withholding info needs to be in an array here, and this has to do with handling split residency / multiple state returns. 

But it does seem like the state wages fields should always be filled in any case, so I'll add those.

thegrims (approved):

PR #836, opened 2022-01-04
Automerge dependabot minor

thegrims:
- Feature: Auto merge Dependabot minor and patch version PRs to reduce PR approval overhead


.github/workflows/automerge-dependabot-minor.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1.1.1
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Enable auto-merge for Dependabot PRs
+        if: ${{contains(steps.metadata.outputs.dependency-names, 'my-dependency') && (steps.metadata.outputs.update-type == 'version-update:semver-patch' || steps.metadata.outputs.update-type ==  'version-update:semver-minor')}}
```
zakpatterson, line 19: where does `my-dependency` come from?

zakpatterson (commented):
Can we find a way to conditionally merge this depending on passing the lint and test workflows? For example if prettier/eslint/eslint-plugins have a version update that changes a parsing rule, then we would want that PR to not be merged so we can amend it manually?

zakpatterson:
Looking around, it seems you can put multiple jobs in a single workflow, and use the `needs` keyword on a job to make it depend on other jobs. So maybe we should merge `lint`, `test`, `build` and this merge workflow into a single workflow file?

.github/workflows/automerge-dependabot-minor.yml
thegrims: removed!

thegrims:
> Looking around, it seems you can put multiple jobs in a single workflow, and use the `needs` keyword on a job to make it depend on other jobs. So maybe we should merge `lint`, `test`, `build` and this merge workflow into a single workflow file?

Done! Let me know what you think :)

zakpatterson (approved):
Looks good!

PR #843, opened 2022-01-11
State migrations for hydration and loading files

noahm:
- Bugfix

This uses the built-in migrations feature of `redux-persist` to migrate the state shape to the latest whenever it is hydrated. (I believe the initial code from #130 may have never worked.) It also uses a shared function to do the same when loading from a JSON file so older files will continue to function with upgrades to the app.

The migrations feature of `redux-persist` depends on a versioning scheme which is a bit overkill for where this app is at the moment, but I think it will be useful in the long run. We just need to be vigilant about bumping the version and including a migration step for each change to the state interface.

(Also fixes a typo in the menu.)

*Maintainer note*:
Fixes an error thrown if localStorage is missing parameters currently expected, due to version updates. Reproducing on master:
```ts
let store = JSON.parse(window.localStorage["persist:root"])
let y2020 = JSON.parse(store.Y2020)
y2020.healthSavingsAccounts = undefined
store.Y2020 = JSON.stringify(y2020)
window.localStorage["persist:root"] = JSON.stringify(store)
```
Will throw an error after switching to Health Savings Account tab.

### DIFF
--- a/src/redux/fs/FSReducer.ts
+++ b/src/redux/fs/FSReducer.ts
@@ -12,7 +12,7 @@ import { FSAction } from './Actions'
  * It will overwrite whatever state exists with whatever
  * state it finds, but needs none of its own state.
  */
-const fsReducer = <S, A extends AnyAction>(
+const fsReducer = <S extends YearsTaxesState, A extends AnyAction>(
   filename: string,
   reducer: Reducer<S, A>
 ): Reducer<S, A & FSAction<S>> => {
@@ -23,7 +23,7 @@ const fsReducer = <S, A extends AnyAction>(
       case 'fs/recover': {
         return {
           ...newState,
-          ...migrateEachYear(action.data as unknown as YearsTaxesState)
+          ...migrateEachYear(action.data)
         }
       }
       case 'fs/persist': {
--- a/src/redux/fs/FSReducer.ts
+++ b/src/redux/fs/FSReducer.ts
@@ -1,7 +1,7 @@
 import { AnyAction, Reducer } from 'redux'
 import { download } from '.'
-import { YearsTaxesState } from '..'
 import { migrateEachYear } from '../migration'
+import { USTPersistedState } from '../store'
 import { FSAction } from './Actions'
 
 /**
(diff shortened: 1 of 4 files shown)


src/redux/store.ts
```diff
@@ -5,43 +5,41 @@ import {
...
+/**
+ * Each key of these migrations will only run if it is less than the current version.
+ * Even though this is a general purpose migration function that should handle most changes,
+ * a new key will need to be added here with each change to the state shape.
+ **/
+const migrate = createMigrate({
+  '0': (state) => {
+    // this type dance is unfortunate, but createMigrate isn't typed generically...
+    const migrated = migrateEachYear(state as unknown as YearsTaxesState)
+    return migrated as unknown as typeof state
```
zakpatterson, line 29: I bet you can get rid of this. I agree I don't like the type in `redux-persist` could probably be better, but it's just requiring that the parameter and return is a subtype of `PersistedState`. So you can enforce any union type of `PersistedState` in the migration functions, see comment in `migration.ts`.

src/redux/fs/FSReducer.ts
```diff
@@ -21,7 +23,7 @@ const fsReducer = <S, A extends AnyAction>(
       case 'fs/recover': {
         return {
           ...newState,
-          ...action.data
+          ...migrateEachYear(action.data as unknown as YearsTaxesState)
```
zakpatterson, line 26: Could probably get rid of this double cast

zakpatterson (changes requested):

src/redux/store.ts
noahm: Found a much cleaner way by dropping their whole system for a plain async function with just our simple logic.

src/redux/migration.ts
```diff
@@ -0,0 +1,19 @@
...
+import { TaxYears } from 'ustaxes/data'
+import { YearsTaxesState } from '.'
+import { blankState } from './reducer'
+
+/**
+ * Ensures all default fields are present on each year's data
+ * @returns state, mutated
+ */
+export function migrateEachYear(state: YearsTaxesState) {
+  for (const year of enumKeys(TaxYears)) {
```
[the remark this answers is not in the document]
noahm: I don't see it needing any reference to `PersistedState` for this function specifically (it doesn't care about that part of state at all). But the rest of it sounds great.

noahm:
@zakpatterson I think the latest commit addresses all that feedback. Let me know if you'd prefer to have it rebased.

zakpatterson:
Nice find on the utilitytype

zakpatterson (approved):
Nice work!

Issue #848, opened 2022-01-13
Add support for uncovered transactions (other investment income) F8949


zakpatterson:
This covers a lot of things people might have, such as real estate sales, cryptocurrency transactions.

* [ ] Add UI to add positions one at a time like other forms (Dependents, W2s, 1099s)
* [ ] Add UI to add investment positions in bulk via csv parsing

PR #851, opened 2022-01-14
Support unreported capital transactions (Form 8949)

zakpatterson:
Adds a new page to the menu for adding sales of assets not
reported on schedule 1099-B. This can include cryptocurrency,
securities owned since before changes in reporting requirements,
and real estate.

Updates Schedule D to incorporate 8949 data, set up for 2020 and 2021.

Assets live outside of the tax years structure at the root of our state datatype. Assets have optional `closeDate`s, so those assets are selected based on `closeDate` in the current tax year for inclusion on the return.

- Feature


src/components/income/OtherInvestments.tsx
```diff
@@ -0,0 +1,198 @@
...
+          name="name"
+        />
+        <DatePicker
+          maxDate={new Date(2021, 12, 31)}
+          label={positionType === 'Real Estate' ? 'Purchase date' : 'Open Date'}
+          name="openDate"
+        />
+        <DatePicker
+          maxDate={new Date(2021, 12, 31)}
+          label={positionType === 'Real Estate' ? 'Sale date' : 'Close Date'}
```
noahm, line 145: Just as an average person I think "purchase date" (or maybe acquisition date instead) and "sale date" make more sense to me for both Real Estate and Securities than "open/close" dates

src/components/income/OtherInvestments.tsx
```diff
@@ -0,0 +1,198 @@
...
+                />
+                <USStateDropDown label="Property state" name="state" />
+              </>
+            )
+          } else {
+            return (
+              <>
+                <LabeledInput
+                  label="Quantity"
+                  patternConfig={Patterns.currency}
```
noahm, line 170: I don't think Quantity is a currency type? Also, since we're concerning ourselves with quantity we should clarify that the cost of acquisition and sell price are either cost per unit or total amount of the given quantity.

src/components/income/OtherInvestments.tsx
```diff
@@ -0,0 +1,198 @@
...
+        <GenericLabeledDropdown<AssetType>
+          label="Asset Type"
+          name="positionType"
+          dropDownData={['Security', 'Real Estate']}
+          keyMapping={(x) => x}
+          textMapping={show(showAssetType)}
+          valueMapping={(x) => x}
+        />
+        <LabeledInput
+          label="Address"
```
noahm, line 134: "Address" only makes sense for Real Estate. This should just be "Description" or something similar for a Security.

noahm (commented):
Just a few initial thoughts from a quick look. I'd love to have a way to quickly input many transactions but I'll need to think a bit longer on how something like that would need to work.

src/components/income/OtherInvestments.tsx
zakpatterson: You're right. There are some cases where selling happens before buying, but I'll stick to the wording on the 8949 itself, "date acquired" and "date sold or disposed of".

src/components/income/OtherInvestments.tsx
zakpatterson: Ah yes, meant to add a switch there.

src/components/income/OtherInvestments.tsx
zakpatterson: Yeah I'm thinking ahead to where lots can be split. So if I buy 100 units of x and then later sell 50 units, I need to subdivide the initial purchase block...might be confusing. Fixed format in cc8ecec

zakpatterson:
Tracking general conception of this feature on the issue: https://github.com/ustaxes/UsTaxes/issues/848

thegrims:
Minor nitpick, but I think the alignment of the manage assets text is off
image

thegrims:
image
address form should allow numbers for house numbers

thegrims:
Saving real estate gives NaN
image

thegrims:
Saving real estate, opening again, asset type is set to stock

thegrims:
image
Date picker should either not allow sale dates of the non tax year, or show warning that proceeds will not be included in the current tax year filing

zakpatterson:
Fixed display issues for Real estate vs other securities, added some styles

2022-01-19-110412_1477x666_scrot

zakpatterson:
Added warning about closing transactions outside of current year.

2022-01-19-182518_1244x652_scrot

thegrims (approved):

PR #873, opened 2022-01-25
Make github push to netlify on succeeding tests

zakpatterson:
Run all tests and build as github actions, then only on success, deploy preview to netlify.

- Build-related changes


artburkart:
Based on the results of that final GitHub Action invocation, I wonder if those environment variables, `NETLIFY_AUTH_TOKEN` and `NETLIFY_APP_ID`, [were maybe not set](https://github.com/ustaxes/UsTaxes/settings/secrets/actions) at the time.


GitHub Action output

PR #881, opened 2022-01-31
Fix box 12 formatting and incorrect medicare wages assumption

### LINKED ISSUE #879: Uncaught (in promise) Error: Validation Failed: /box12/DD: must be number || Numeric value passed as string
Looking at the console, it looks like the value is stored as a string. The other fields, such as income, show as a number. ​

Numbers changed for reasons
income: 00000.00, medicareIncome: 000000.00,

as opposed to

box12: {
...
DD: "0000.00"
...
}

### LINKED ISSUE #880: Medicare wages are not allowed to be greater than wages
I have an error saying "Medicare wages are not allowed to be greater than wages" on the review and print section. This is definitely wrong because my W2 box 5 is larger than my box 1. This is allowed since anything contributing to pre-tax things like 401(k)s (box 12D). I can verify this by seeing that Box 1 + Box 12D = Box 5.

zakpatterson:
- Bugfix

Fixes #879 
Fixes #880 

Thanks to @dweiss2 and @gnoejuan for their clearly explained bug reports and help figuring out the causes of these issues!

#879: We were not parsing data from string to number and back again for the user form with respect to all the box 12 fields. This caused attempts to save user input strings where numbers were expected.

#880: We were assuming that medicare wages are less than or equal to total wages on the W2. This is a result of incompletely implementing the W2 box 12 data, as box 12 can contain information such as elective 401k contributions. These contributions are not included in Box 1 wages so no income tax is charged on them, but they are included in Box 5 medicare wages because medicare tax may be collected on them. 

This also fixes a no key specified error on the box 12 fields, and a material ui error where the wrong kind of container was used to subdivide the box 12 fields.

### DIFF
--- a/src/components/income/W2JobInfo.tsx
+++ b/src/components/income/W2JobInfo.tsx
@@ -1,4 +1,5 @@
 import { ReactElement, ReactNode, useState } from 'react'
+import _ from 'lodash'
 import { useDispatch, useSelector, TaxesState } from 'ustaxes/redux'
 import { Helmet } from 'react-helmet'
 import { FormProvider, useForm, useFormContext } from 'react-hook-form'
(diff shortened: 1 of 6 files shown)


thegrims (approved):
LGTM! Thanks for the fix

PR #883, opened 2022-02-01
Fix w2 form to accept any employer name

### LINKED ISSUE #882: Slight Disagreement with employer name validation, not sure if it matters though
I worked for a Texas grocery chain, H-E-B, LP

The form won't submit with the validation message: "Input should only include letters and spaces"

Does it matter that on the W-2 its "H-E-B, LP", which I **assume** means that's the legal name of the company, or does it make no difference?

zakpatterson:
Employer names are just informational for the user, and should have no restriction as to format.

Also updates testing so that arbitrary names are generated for employer names.

Fixes #882 

- Bugfix

### DIFF
--- a/src/components/income/W2JobInfo.tsx
+++ b/src/components/income/W2JobInfo.tsx
@@ -222,7 +222,7 @@ export default function W2JobInfo(): ReactElement {
         <LabeledInput
           autofocus={true}
           label="Employer name"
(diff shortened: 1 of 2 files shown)


zakpatterson:
This is a very small change, I reviewed my own PR for efficiency.

PR #884, opened 2022-02-02
changing dependabot config to combine major and minor updates

thegrims:
- Bugfix

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,17 +12,6 @@ updates:
       interval: "daily"
     target-branch: "master"
     ignore:
-      - dependency-name: "*"
-        update-types: ["version-update:semver-major"]
-    labels: 
-      - "minor"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -6,17 +6,6 @@
...
-    target-branch: "master"
-    ignore:
-      - dependency-name: "*"
-        update-types: ["version-update:semver-major"]
-    labels: 
-      - "minor"
-  # major updates:
   - package-ecosystem: "npm"
     directory: "/" # Location of package manifests
     schedule:
```
zakpatterson, line 11: Should we just get rid of the package version locks here even if each generated PR has to remain open for a while?

thegrims: Sure!

zakpatterson:
Should be fine, guess there's no way to know for sure except...

PR #903, opened 2022-02-05
Unified ira form

ctSkennerton:
While attempting to implement the IRA tax deduction I realised that IRA contributions were not being captured anywhere. This change creates a new UI page for IRAs that lets a user input both contributions and distributions from an account.

A couple of things that I'd like feedback on:
- I moved the IRA information under their own key in the Information object. Is this the right approach? I could also see creating an umbrella key like "savingsAccounts" and have IRAs and HSAs underneath that. Something like this: 
    ```
    Information: {
        savingsAccounts: {
            hsas: [],
            iras: []
    }
    ```
- Previously only IRA distributions were captured under the `f1099s` in `Information`. How should the older data be migrated?

- Feature


zakpatterson:
So an IRA could produce a 5498 (for contributions) **and** a 1099-R (for distributions), right? In one year someone could receive either a 5498 or a 1099-R or both for the same retirement account?

Is there something special about health savings accounts that is not present on 1099-SA?

At this point we're looking to support 1099-{B, INT, DIV, SSA, SA, R}. So it seems to me that all of these 1099s could be treated similar to each other in the webforms, and they could all be siblings of each other in the Information data structure.

I think as much as possible there should be a one-to-one correspondence between the tax forms the user receives, to the webforms. We should maintain as much as possible a simple pipeline from `TaxForm -> WebForm -> Information state object -> Return form classes -> PDFs`  so that we only have to conceptualize two adjacent segments of that pipeline at a time.

An IRA / HSA is otherwise no different from a brokerage account. Instead of modeling Information based on what accounts the user actually has, we're presently modeling it based on what 1099s/W2s the user actually receives. This is more complicated by the unreported 8949 income, where we do have a reason to track assets from year to year since we might get data about assets that have not been sold in the current tax year.

But I think in general the information object should mirror the tax forms (or transaction reports) that a user could have for a given tax year, with as little interpretation as possible. All the interpretation should happen in the return form classes step.

ctSkennerton:
> So an IRA could produce a 5498 (for contributions) and a 1099-R (for distributions), right? In one year someone could receive either a 5498 or a 1099-R or both for the same retirement account?

correct

> Is there something special about health savings accounts that is not present on 1099-SA?

The coverage type and the dates that you have the coverage are not given on the forms. I made HSA into a single web form since there were some complicated rules around which HSA accounts get counted for a person and how to calculate the maximum contributions based on those two pieces of info. To model that as a separate 1099-SA and 5498-SA would require users to potentially match up those two webforms based on a shared key (like account name) and duplicate the date info. 

My reasoning to extend that to IRAs follows the same logic. However, I do see that there is less complicated logic around IRAs, so I could instead create a 5498 webform.

src/core/data/index.ts
```diff
@@ -222,6 +227,52 @@ export interface HealthSavingsAccount<DateType = string> {
...
+  [IraPlanType.SepIRA]: 'Simplified employee pension (SEP) IRA',
+  [IraPlanType.SimpleIRA]:
+    'Savings incentive match plan for employees (SIMPLE) IRA'
+}
+
+export type IraPlanName = keyof typeof IraPlanType
+
+// Typescript provides no way to access
+// keys of an interface at runtime.
+export const iraPlanNames: IraPlanName[] = [
```
zakpatterson, line 249: Generally if you need an array of the names of an enum, you can use [enumKeys](https://github.com/ustaxes/UsTaxes/blob/7056ea37276f6d610b9dee5d9b1e7847871be82a/src/core/util.ts#L8).

src/core/tests/arbitraries.ts
```diff
@@ -451,6 +459,68 @@ export class Arbitraries {
         })
       )
 
+  ira = (): Arbitrary<types.Ira> =>
+    fc
+      .tuple(
+        words, //payer
```
zakpatterson, line 465: you can just make this an arbitrary string, no reason to restrict to alpha chars.

src/components/savingsAccounts/IRA.tsx
```diff
@@ -0,0 +1,281 @@
...
+          }
+          name="personRole"
+          keyMapping={(p: Person, i: number) => i}
+          textMapping={(p: Person) =>
+            `${p.firstName} ${p.lastName} (${formatSSID(p.ssid)})`
+          }
+        />
+        <LabeledInput
+          name="contributions"
+          label="Contributions"
```
zakpatterson, line 184: Check out [`boxLabel`](https://github.com/ustaxes/UsTaxes/blob/323fa4716c8ae24e85a8e21052425eb17c92ee1c/src/components/income/W2JobInfo.tsx#L230), it'd be nice to explicitly map these to box numbers from the 1099 or 5498

zakpatterson (commented):
know you're still working on this, just drawing a few things to your attention.

zakpatterson:
> thank you both for looking over this! Since this modifies the Information object, how do data migrations work?

I think the existing migration [here](https://github.com/ustaxes/UsTaxes/blob/d70b5813a97eebd139de0886774d965afc3d0121/src/redux/migration.ts) will be enough. Basically that function ensures that all fields are present for every tax year, and then plays existing keys from localstorage on top. So the migration will take the valid, blank state which will contain your individualRetirementAccounts key, then load localStorage on top of that. 

I think as long as we're just adding keys to Information we don't have to do anything, but if we ever want to move keys around we'll have more work to do

ctSkennerton:
> I thought that the 1099 page would handle 1099-R for employer pension plans, where the taxpayer doesn't have an associated retirement account. What is the difference between a 401(k) and an IRA from our point of view here? Don't they both involve contributions and distributions?

From looking at the 1040 things are broken down by account type. 401(k), 403(b), and annuities are in one box and then the various flavors of IRA in another box. Other forms like IRA deduction care about the type of IRA. I haven't seen anything that changes based on whether the IRA is an employer pension plan. From that perspective 401(k) is the employer pension plan that is being left on the 1099-R form. IRAs have form 5498 to track the contributions but I don't think there is a corresponding form for a 401(k)

zakpatterson:
>  IRAs have form 5498 to track the contributions but I don't think there is a corresponding form for a 401(k)

Yep that makes sense. Ok!

zakpatterson (approved):

src/components/savingsAccounts/IRA.tsx
```diff
@@ -0,0 +1,300 @@
...
+          <br />
+          contribution: <Currency value={toIra(ira).contributions} />
+          <br />
+        </span>
+      )}
+    >
+      <Grid container spacing={2}>
+        <LabeledInput
+          name="payer"
+          label="payer for this account"
```
thegrims, line 159: Minor nitpick, I think payer should be capitalized to fit the format of the rest of the form fields on the IRA page

thegrims (approved):
LGTM! Thanks for your contribution!

PR #932, opened 2022-02-16
Get ts-jest, for easier debugging setup

zakpatterson:
- Build-related changes

Just using craco test made it difficult to set up vscode with breakpoints. I'm not exactly sure why, but switching to ts-jest made it easier for me to set up breakpoints.

Using this, I can reliably auto-attach to the node testing process in vscode, and set breakpoints from running tests.

Also, setting up tests this way revealed several type errors in our tests.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,10 @@
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
+module.exports = {
+  preset: 'ts-jest',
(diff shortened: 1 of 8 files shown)


zakpatterson:
I don't think this is needed anymore, I was finally able to get vscode breakpoints to work as well as the breakpoints to be connected to the browser with our current setup, although I think we should document this for contributors.

PR #938, opened 2022-02-19
Use Fillable Schedule B

ssigwart:
- Bugfix

Schedule B isn't currently filled in because the PDF wasn't the fillable version.


thegrims:
Good catch, thanks!

PR #940, opened 2022-02-21
Add form 6251 (AMT) for Stock Options (Form 3921)

ssigwart:
- Feature

This adds a place to enter form 3921 for stock options and then form 6251 to compute AMT.  I'm not totally sure if the logic for Part III is correct, particularly around Schedule D (e.g. Line 20).

Form 6251 instructions are at https://www.irs.gov/instructions/i6251


ssigwart:
I fixed the error.  I'm not sure if the options should be tied to a particular person.  I'll look into that.

PR #974, opened 2022-03-06
2021 tax rate fixes and tests

ssigwart:
- Fixes taxes for under $100,000.
- Adds tests to confirm amounts.

See https://www.irs.gov/instructions/i1040gi#idm140354391131776

- Bugfix


src/forms/Y2021/tests/taxRates.test.ts
```diff
@@ -0,0 +1,253 @@
...
+import { CURRENT_YEAR } from '../data/federal'
+import { computeOrdinaryTax } from '../irsForms/TaxTable'
+
+describe('Tax rates', () => {
+  it('test should be updated for new year', async () => {
+    // WARNING! Do not just chagne the year. Also update the expected tax amounts below!
+    expect(CURRENT_YEAR).toEqual(2021)
+  })
+  it('ordinary taxes for single status should be correct', async () => {
+    expect(Math.round(computeOrdinaryTax(FilingStatus.S, 5))).toEqual(1)
```
zakpatterson, line 11: This could be handled a bit better I think, no reason you can't just have a CSV file loaded, loop over it and assert `f(x) = y` for each row in the file for example, or even doing a `.foreach` just on a JS array loaded above here would be fine.

zakpatterson (changes requested):

src/forms/Y2021/tests/taxRates.test.ts
zakpatterson: There's also a property based testing approach that could work. You could have the tax table you linked imported directly as a CSV, and then generate arbitrary incomes, and assert for income bracket I and tax T, 

```
I_0 < x <= I_1, T_0 < f(x) <= T_1`
```

ssigwart: I updated it to pull in a CSV, then a few points in every tax table range.

thegrims (approved):
LGTM! Thanks for fixing this

tmm1:
> See https://www.irs.gov/instructions/i1040gi#idm140354391131776

how did you convert this to csv?

ssigwart:
> > See https://www.irs.gov/instructions/i1040gi#idm140354391131776
> 
> how did you convert this to csv?

If I remember correctly, I copied it into VS Code, then converted all the tabs to commas and used some regex searches to remove unnecessary lines.

PR #982, opened 2022-03-08
Migrate IL forms for 2021, fix filing status

### LINKED ISSUE #555: IL Tax return can only select single filing status
The filing status field is implemented as a radio button. It's not clear how to fill this field with PDF lib

### LINKED ISSUE #975: Migrate IL tax for 2021
User from twitter DM'ed the USTaxesorg handle asking for us to update the tax forms for 2021.

zakpatterson:
- Feature
- Bugfix 

Fixes #975 

Implements 2021 pdfs as 2020.

Fixes #555  bug where the correct option from a radio group could not be selected (filing status). This adds a new field type RadioSelect, which indicates an index of a number of children checkboxes to select in a radio group.


src/core/pdfFiller/fillPdf.ts
```diff
@@ -19,8 +20,24 @@ export function fillPDF(pdf: PDFDocument, fieldValues: Field[]): PDFDocument {
...
+      const setValue = children[value.select].getOnValue()
+      if (setValue !== undefined) {
+        pdfField.acroField.dict.set(PDFName.of('V'), setValue)
+        children[value.select].setAppearanceState(setValue)
+      } else {
+        console.error(children)
+        throw new Error(
+          `Error handling RadioGroup, could not set index ${value.select}`
+        )
+      }
```
thegrims, line 39: What is the advantage of this over checking and filling pdf radio groups using the radio group api directly? For example this is what is being used for the OR-40 pdf radio groups

```
else if (pdfField instanceof PDFRadioGroup) {
   try {
        pdfField.select(value?.toString() || '')
```

zakpatterson: We need more tooling around testing buggy pdfs, but the pdffield in the IL form—even though it behaves as a radio group with clicks—is of type PDFCheckbox. I eventually figured out that the widget itself has children. But iterating over the list of fields would never detect them. 

So the advantage is now we have two ways to select radio buttons depending on how they were implemented by the poor person that has to make these forms.

zakpatterson: I will update the comments to indicate it’s for cases where the widget is implemented this way. I hadn’t come across a field passing instanceof PDFRadioGroup yet.

thegrims: Interesting, it's a bummer the radiogroup isn't used in the IL form

zakpatterson: I don't know pdf well enough. Somehow the field behaves as a radio group but for some reason pdflib doesn't detect it as such. But that link in my comment shows other people have run into this too.

thegrims (approved):

PR #995, opened 2022-03-11
fix menu header on settings page

tmm1:
- Bugfix

Previously said "undefined - undefined"


zakpatterson (changes requested):
Your solution is fine, just want to point out a couple things:

* This getTitleAndPage function is only ever called one time, from within this same file, so it  doesn't need that sections argument it should just be using the imported variable IMO.
* `drawerSections` contains data about each page that's displayed on the left side bar, but this variable includes data about all the pages that are not displayed on the left side bar:

https://github.com/ustaxes/UsTaxes/blob/88bf134f31da96b3ca3b962f0aa41b4a77d9616c/src/components/Main.tsx#L92

So, I would recommend fixing it by incorporating the backpages array.

tmm1:
> So, I would recommend fixing it by incorporating the backpages array.

Done

thegrims (approved):

PR #997, opened 2022-03-11
fix invisible menu on iPad safari

tmm1:
- Bugfix

use `isMobileOnly` which excludes tablets


thegrims:
@tmm1 is the menu not showing up at all on the iPad or is it the hamburger swipeable style mobile menu?

PR #1001, opened 2022-03-11
support k1 (form 1065) box 5 interest income

tmm1:
cc @ssigwart

### DIFF
--- a/src/redux/migration.ts
+++ b/src/redux/migration.ts
@@ -16,6 +16,13 @@ export const migrateEachYear = <S extends USTState>(state: S): S =>
       }
     })
 
+    // Ensure interestIncome on K1s
+    acc[year].scheduleK1Form1065s.forEach((k1) => {
+      if (k1.interestIncome === undefined) {
+        k1.interestIncome = 0
+      }
+    })
+
     return {
       ...acc,
       [year]: {
--- a/src/components/income/PartnershipIncome.tsx
+++ b/src/components/income/PartnershipIncome.tsx
@@ -38,6 +38,7 @@ interface ScheduleK1Form1065UserInput {
   isForeign: boolean
   isPassive: boolean
   ordinaryBusinessIncome: string
+  interestIncome: string
   guaranteedPaymentsForServices: string
   guaranteedPaymentsForCapital: string
   selfEmploymentEarningsA: string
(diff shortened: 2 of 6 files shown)


ssigwart (commented):
I don't know exactly how box 5 is handled, but this code looks reasonable to me.

src/redux/migration.ts
```diff
@@ -16,6 +16,13 @@ export const migrateEachYear = <S extends USTState>(state: S): S =>
       }
     })
 
+    // Ensure interestIncome on K1s
+    acc[year].scheduleK1Form1065s.forEach((k1) => {
```
zakpatterson, line 20: I didn't notice it in the w2s part above, but this is modifying the state object in place, and then returning it. I think in redux this will cause a rerendering problem if there is something depending on a subscription of the `scheduleK1Form1065s` array. Is that right?

zakpatterson (changes requested):

src/redux/migration.ts
zakpatterson, line 20: I didn't notice it in the w2s above, but there is state modification in place happening here, which is not good. It might be ok at this stage, since this is all only happening at first load, but in general in redux it leads to errors in rerendering caused by subscribers not being updated.  The issue is basically this:

```js
const ks = ['root']

const oldState = {
  root: [{ child: 'child1' }]
}

const newState = ks.reduce((acc, key) => {
  acc[key].forEach((item) => {
    item.child += ' hello'
  })
  return {
    ...acc
  }
}, oldState)

newState == oldState
// false -- Good, if a subscriber is looking at the root state, it rerenders

newState.root === oldState.root
// true -- Bad, a subscriber looking at the child of the state does not rerender

```

zakpatterson (changes requested):

src/redux/migration.ts
tmm1: Hm, okay. I assumed it was okay here since the intent is to migrate data. I don't have much experience with redux and was going off https://github.com/ustaxes/UsTaxes/pull/955#discussion_r819250361 and the surrounding code.

zakpatterson: Yea I agree, if we need to fix it later will do all cases in a different pr. Thanks!

zakpatterson (approved):

PR #1003, opened 2022-03-13
Switch csv-parse for papaparse

zakpatterson:
csv-parse is causing too many issues with being based as a node.js project and polyfilling for browser, papaparse seems to be simpler.

### DIFF
--- a/src/components/income/TransactionImporter.tsx
+++ b/src/components/income/TransactionImporter.tsx
@@ -125,16 +125,11 @@ export const TransactionImporter = (): ReactElement => {
     setFieldAssignments(newFieldAssignments)
   }
 
-  const onHandle = async (contents: string): Promise<void> => {
+  const onHandle = (contents: string): void => {
     setRawContents(contents)
     run(preflightCsv(contents)).fold(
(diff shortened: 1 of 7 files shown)


src/components/income/TransactionImporter.tsx
```diff
@@ -126,11 +127,15 @@ export const TransactionImporter = (): ReactElement => {
 
   const onHandle = async (contents: string): Promise<void> => {
     setRawContents(contents)
-    const res = await preflightCsv(contents).catch((parseError) => {
-      console.error("Couldn't parse CSV", parseError)
-      return []
-    })
-    setPreflightTransactions(res)
+    run(preflightCsv(contents)).fold(
```
zakpatterson, line 130: both of these should be void

zakpatterson: Hm I can't "request changes" on my own PR.

zakpatterson:
Errors are not showing if there's an unparsable row...fixing.

thegrims (approved):
LGTM

PR #1017, opened 2022-03-17
Allow Schedule B to be attached multiple times

zakpatterson:
- Bugfix

We've been seeing errors on most PRs now, due to generated taxpayer information containing more sources of interest received than fit on Schedule B. 

Solving this by attaching schedule B numerous times.


zakpatterson:
I tested this out, reviewing my own work so we can merge the dependency updates.

Issue #1032, opened 2022-03-21
App is broken on Node 17


xxxserxxx:
None of the instructions work on my machine:

Running with npm:
```sh
10043(master) » git log -1 | head -n 1
commit 2d2d6f8254fb35fbe82a6c7173a50d97287e842e
10044(master) » npm install
npm ERR! semver.simplifyRange is not a function

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/ser/.npm/_logs/2022-03-21T21_42_15_479Z-debug-0.log
```

Said log is 6,816 lines long and mostly "silly reify" and "verbose unfinished npm timer" messages. 

Running with docker-compose; the "build" exits with status 0 and showed no obvious errors, but the run:

```sh
10048(master) » docker-compose up
...
> ustaxes@0.1.19 start
> node ./notice.js && ts-node ./scripts/env.ts && ts-node ./scripts/setup.ts && concurrently npm:dev
... (bunch of copyright wording)
[Error: EACCES: permission denied, open 'src/core/data/validation.json'] {
  errno: -13,
  code: 'EACCES',
  syscall: 'open',
  path: 'src/core/data/validation.json'
}
exit code: 1
```

Ok, how about the desktop:

```sh
10052(master) » npm run desktop

> ustaxes@0.1.19 desktop
> tauri dev

sh: line 1: tauri: command not found
```

What's `tauri`? It isn't listed in the dependencies.

- rustc --version
  rustc 1.60.0-nightly (e7aca8959 2022-02-09)
- cargo --version
  cargo 1.60.0-nightly (c082648 2022-02-08)
- npm --version
  8.5.4
- Linux
  Linux 5.16.14-arch1-1

zakpatterson:
Hi, thanks for trying out the project, I'll try to help you get it sorted.

Up until now I think `nvm use stable` has been installing a node 16 for me, but now it seems it's providing a node 17. Our project does seem broken on node 17 although I'm not seeing the same behavior as you.

The Dockerfile is also trying to load latest node, so that should explain that.

So I suggest first trying to get on the node version that we're using in our CI, which is node 16.

Assuming you have nvm, if not [install here](https://github.com/nvm-sh/nvm#install--update-script)

Are you using direnv? Here is a good config to put in our project directory:

.envrc:
```
export NVM_DIR="$HOME/.nvm"

. "$NVM_DIR/nvm.sh"  # This loads nvm
[...]

PR #1033, opened 2022-03-22
Improve startup docs

zakpatterson:
Addressing feedback from #1032

* Delegates all desktop setup to tauri's official docs
* Specifies to develop locally on node 16 only, and gives some guidance on setting that up
* Moves setup instructions out of the Architecture doc
* Fix docker-compose configuration, which could cause an error if the user doesn't already have a node_modules folder prior to running.

- Docs

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -2,33 +2,6 @@
 
 Thank you for your interest in this project. The below should summarize the general design framework used in this project, and hopefully guide you in understanding the code and getting ready to contribute to the project. If anything at all is unclear please join the discord linked on the readme and feel free to ask any questions you might have.
 
-## Stack and developer requirements
-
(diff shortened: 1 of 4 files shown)


docs/ARCHITECTURE.md
```diff
@@ -2,33 +2,6 @@
...
-Running the desktop version requires at least the following to be installed:
-
-```
-libssl-dev
-libgtk3-dev
-libgtk-3-dev
-libsoup2.4-dev
-webkit2gtk-driver
-libwebkit2gtk-4.0-dev
-```
```
thegrims, line 30: Do we need to keep this part of the instructions or is this no longer the case for Linux?

zakpatterson: Still required, I’m just delegating to the tauri setup doc linked in the new get started section of the readme.

thegrims (approved):
LGTM, thanks for updating the docs

PR #1040, opened 2022-03-24
Enhanced form parser UI (replacement for npm formgen script)

zakpatterson:
I'm working on a fast form field filler generator machine...

So far it can: 

* Insert a PDF from the file system
* Render it side by side with a list of fields
* Fill all the fields of the PDF either with their field names or field indices (from `getForm().getFields()`)
* On selecting a field from the list on the left, the corresponding field will be highlighted on the right.

Goals in mind:
* Digest all of our available forms this way
* Create a user-definable encoding for what should go in a field
* Create a power-user mode where the code for the given form is editable in a browser sandbox? [imagining like this](https://www.solidjs.com/tutorial/introduction_basics), but the code needs to be savable.

This editor could just be its own application, but as implemented now we could have user data available to the pdf render.

This is far from mergeable, just looking for feedback and comments, of course PRs to this branch are welcome too if there is interest.


thegrims:
```
errors.ts:102 Uncaught (in promise) Error: Attempted to set text with length=2 for TextField with maxLength=1 and name=or-40-n-p1-14
    at new ExceededMaxLengthError (errors.ts:102:1)
    at PDFTextField.setText (PDFTextField.ts:154:1)
    at Form.tsx:165:1
    at Form.tsx:153:1
    at Array.forEach (<anonymous>)
    at fillAllFields (Form.tsx:153:1)
```

Getting this error when running fill with indices on or-40

thegrims:
I really like the feature of highlighting the field that corresponds to a field name, that will be really useful

zakpatterson:
> I really like the feature of highlighting the field that corresponds to a field name, that will be really useful

Just wish I could make it go the other way around. There's no way to get coordinates of fields. 

I could draw red boxes on fields one at a time, then render the pdf for each field, find the red box's coords, and overlay clickable buttons. Maybe eventually I'll convince myself it's worth doing.

PR #1048, opened 2022-03-27
Fix for excess FICA taxes (Schedule 3 line 11)

ctSkennerton:
FICA taxes are calculated per person which means that any excess amount that can be claimed should be calculated for the primary and the spouse separately. The previous logic did not differentiate the w2s between individuals so the excess amount was calculated to be too high.

- Bugfix


thegrims:
Could this fix also be added for the 2020 schedule 3 file and tests?

ctSkennerton:
sure, done.

zakpatterson (approved):

PR #1049, opened 2022-03-27
Fix two field issues with Schedule 1

### LINKED ISSUE #1047: Incorrect Schedule 1 Generation
Part II section 24

Line item 26 calculation is placed in line item 25

Version 1.19

ctSkennerton:
- Fixes #1047
- fix issue where 'HSA' would be added to line 8z even if there
  was no additional income. Caused because form 8889 line 20 always
  returns a number but was only checking for undefined. It will
  be undefined only if the form itself is not attached.

- Bugfix


thegrims (approved):
LGTM, thanks for the bugfix!

thegrims:
Is this also an issue on the 2020 schedule 1 form?

PR #993, opened 2022-03-10
Remove references to yarn

tmm1:
support was removed in dab49f99e819e67ffc3e0181be39d97dc8cce6cd

**What kind of change does this PR introduce?** (delete not applicable)
- Bugfix

PR #1087, opened 2022-04-05
add support for yarn back

tmm1:
partial revert of #508

cc #993

- Bugfix
- Feature
- Docs
- Code style update
- Refactor
- Build-related changes
- Other, please describe:

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,0 @@
-//.npmrc file
-engine-strict = true
--- a/package.json
+++ b/package.json
@@ -110,8 +110,5 @@
     "transformIgnorePatterns": [
       "/node_modules/(?!@tauri-apps)"
     ]
-  },
-  "engines": {
-    "yarn": "please-use-npm"
   }
 }


thegrims:
I think the reason we added a restriction to NPM only is users were having issues with installing with yarn with the NPM infrastructure set up. I'm fine with using yarn instead of NPM, but I think we should pick one or the other to work with to lower the amount of dev environment issues need support.

zakpatterson:
I think `yarn` agrees here. If I apply this change and then run `yarn install`, 

```
yarn install v1.22.18
info No lockfile found.
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
[1/4] Resolving packages...
```

I think that projects that support both yarn and npm will typically gitignore the lock files, [for example tauri](https://github.com/tauri-apps/tauri/search?q=package-lock).

But if we do that, then `npm ci` won't work anymore and resolution will take a very long time. I think this adds more friction for devs than having to use `direnv` or just remember npm. 

Is there anything else we can do to enable both? Or is yarn more popular /convenient so we should just switch?

zakpatterson:
Just going to close this, but if anyone feels strongly about adding support for yarn, feel free to comment. Thanks

PR #1095, opened 2022-04-08
Fix 8812 Math #1093

### LINKED ISSUE #1093: Schedule 8812
Schedule 8812 line 5 should be 0 if line 4a is 0.

Part I-B is blank, but I think that is a result of the first problem.  

Part III line 33 is 40,000. It should be 60,000 for married filing jointly.  
Line 34 should be line 3 minus line 33. Instead it is equal to line 3.

zakpatterson:
Fixes several math bugs in 8812, and refactors common test code for reusability.

Fixes #1093

- Bugfix

### DIFF
--- a/src/core/util.ts
+++ b/src/core/util.ts
@@ -168,3 +168,19 @@ export const parseFormNumberOrThrow = (x: string | undefined): number => {
   }
   return res
 }
+
+export const nextMultipleOf =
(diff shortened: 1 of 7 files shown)


thegrims (approved):
LGTM, thanks for the bugfix!

PR #1096, opened 2022-04-08
Age and blindness

ctSkennerton:
Add in Age and blindness status to the front end and the data model. I needed the date of birth for doing the California taxes (which asks for the date of birth for everyone) so I added in the date to the Person interface. This caused some complications though because AJV doesn't like Date objects so for the Information object I serialize to string. I tried to make Person a generic interface that could take a date or a string but that doesn't seem to work well when you extend a generic interface. 

I settled on enriching the `TaxPayer` class so that it would use the version with the `Date` object while the person interface in `Information` uses the string version. As part of this I also went through and used the `TaxPayer` class in many cases where there was a call to the `Information` object.

Not sure that this is the best way to handle this so would be open to other suggestions.

- Add the `TaxPayer` to form f1040
- clean up the constructors of many of the schedules as they were
  passing in redundant information or would construct new `TaxPayer`
  objects
- Remove unnecessary calls to create new `TaxPayer` objects when generating
  the fields of a form. In many cases this object was available
  already as part of f1040.
- Convert Dependent logic to use the `dateOfBirth` field and remove the `birthYear` field

- Feature
- Refactor

### DIFF
--- a/src/components/TaxPayer/PersonFields.tsx
+++ b/src/components/TaxPayer/PersonFields.tsx
@@ -7,7 +7,12 @@ import {
   ListItemSecondaryAction
 } from '@material-ui/core'
 import { useDispatch, useSelector, TaxesState } from 'ustaxes/redux'
-import { formatSSID, LabeledInput } from 'ustaxes/components/input'
+import {
+  formatSSID,
+  LabeledInput,
+  LabeledCheckbox,
+  DatePicker
+} from 'ustaxes/components/input'
 import { Patterns } from 'ustaxes/components/Patterns'
 import { removeDependent } from 'ustaxes/redux/actions'
 import { Person } from 'ustaxes/core/data'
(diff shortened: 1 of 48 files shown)


zakpatterson:
Great! 

For AJV, we're using a generator that creates an AJV schema using our typescript code, but I've seen that it just completely ignores generics. So I've settled on a pattern like this:

```
interface TaxPayer<DateType = Date> {
   ...,
   birthDate: DateType
}
type TaxPayerString = TaxPayer<String>
```

And the generator will generate for TaxPayerString. Then I think that type can be used where needed to get a validator.

ctSkennerton:
Okay I'll try it out. I need to fix up the conflicts, did not see the latest merge, which is great by the way!

ctSkennerton:
closing this pull request in favor of #1098

PR #1098, opened 2022-04-08
Age and blindness redux

ctSkennerton:
Redo of #1096 after taking into account the suggestion that zakpatterson made. Started from new branch to avoid the many merge conflicts with the previous pull request

- Feature


src/core/data/index.ts
```diff
@@ -324,13 +333,15 @@ export interface ContactInfo {
...
   filingStatus?: FilingStatus
-  primaryPerson?: PrimaryPerson
-  spouse?: Spouse
-  dependents: Dependent[]
+  primaryPerson?: PrimaryPerson<D>
+  spouse?: Spouse<D>
+  dependents: Dependent<D>[]
 }
 
+export type TaxPayerDateString = TaxPayer<string>
```
zakpatterson, line 343: I think the build error you're seeing shows that you just need to do this same thing for each of the other types you've modified this way.

src/core/data/index.ts
```diff
@@ -15,24 +15,30 @@ export enum PersonRole {
...
 }
 
-export interface Person {
+export interface Person<DateType = Date> {
   firstName: string
   lastName: string
   ssid: string
   role: PersonRole
+  isBlind: boolean
+  dateOfBirth: DateType
```
zakpatterson, line 24: You'll need to address these changes in a data migration, if you can't make these two fields optional. Otherwise a user that has data in local storage but does not have these fields will have errors.

See: 
https://github.com/ustaxes/UsTaxes/blob/2c3ec7c2ec9d1049d38e927b2894933795748a08/src/redux/store.ts#L79

https://github.com/ustaxes/UsTaxes/blob/2c3ec7c2ec9d1049d38e927b2894933795748a08/src/redux/migration.ts#L10

zakpatterson (changes requested):

src/core/data/index.ts
ctSkennerton: I'm not sure what you mean, I think I've already done that for Person, PrimaryPerson, Spouse, and Dependent.

src/core/data/index.ts
ctSkennerton: Should I add in another function to be run after`migrateEachYear` or do I add into that function?

zakpatterson: I think it would be a new function. The idea with migrations is that each time the state type is changed we add a function which updates the users state to match the expected type. 

So if those functions are called in sequence (chronologically) then hopefully no matter how old someone’s data is it will be corrected to the current type. 

I’m sure we will need a more sophisticated solution soon.

src/core/data/index.ts
zakpatterson: Ah I see. Ok will have to debug  https://github.com/ustaxes/UsTaxes/runs/5882167289?check_suite_focus=true

zakpatterson:
I just had to fix the names in the setup script `scripts/setup.ts` so that the names matched your concrete types. I also added the date type generic to a couple more types.

Hopefully this unblocks you, let me know!

ctSkennerton:
Thank you! I’m reading up about redux-persist to complete the migration part

zakpatterson:
> Thank you! I’m reading up about redux-persist to complete the migration part

Great, I think it should be as simple as wrapping the one migration function we have right now in a new one. A migrate function is just a function from `State => State`. The incoming state object is a runtime object, so its type is just assumed to be a State. 

I'm sure you're figuring this out now, but just in case it saves you a few minutes:

store.ts:
```ts
const migrate = async (state: USTState): Promise<USTState> =>
  migrateEachYear(state)
```

So this could just be:
```ts
const migrate = async (state: USTState): Promise<USTState> =>
  migrateBirthYears(migrateEachYear(state))
```

where `migrateBirthYears` or whatever you call it would be your migration function that adds your two new parameters.

zakpatterson:
I'll fix up the tests - replacing birth year with date of birth.

ctSkennerton:
> > Thank you! I’m reading up about redux-persist to complete the migration part
> 
> Great, I think it should be as simple as wrapping the one migration function we have right now in a new one. A migrate function is just a function from `State => State`. The incoming state object is a runtime object, so its type is just assumed to be a State.
> 
> I'm sure you're figuring this out now, but just in case it saves you a few minutes:
> 
> store.ts:
> 
> ```ts
> const migrate = async (state: USTState): Promise<USTState> =>
>   migrateEachYear(state)
> ```
> 
> So this could just be:
> 
> ```ts
> const migrate = async (state: USTState): Promise<USTState> =>
>   migrateBirthYears(migrateEachYear(state))
> ```
> 
> where `migrateBirthYears` or whatever you call it would be your migration function that adds your two new parameters.

@zakpatterson the redux-persist docs use another mechanism that uses a object which tracks the version of the persisted object and applies only the migrations between the version being rehydrated and the current config.[ I implemented that](https://github.com/ustaxes/UsTaxes/pull/1098/files#diff-b3695b5171edbee1bb22dc8893e3517773bacdd8a51ba2a568e077f6e1e07940R91) before I saw this comment. I can change it if you don't think that is the right way to go but it made sense to me as there will be more migrations going forward to capture data that might have been persisted in different intermediate states.

One thing that is still to be done is that loading from a file [directly calls the migrateEachYear function](https://github.com/ustaxes/UsTaxes/blob/master/src/redux/fs/FSReducer.ts#L35), which means that the data won't come in the right format. But I'm not sure how to go about changing that in a maintainable way. Just changing the function to be the latest migration name will cause the same problem for whatever the next migration is. Any ideas?

zakpatterson:
> @zakpatterson the redux-persist docs use another mechanism that uses a object which tracks the version of the persisted object and applies only the migrations between the version being rehydrated and the current config.[ I implemented that](https://github.com/ustaxes/UsTaxes/pull/1098/files#diff-b3695b5171edbee1bb22dc8893e3517773bacdd8a51ba2a568e077f6e1e07940R91) before I saw this comment.

Ah this is great.

zakpatterson:
I went ahead and merged with master. 

The migrations issue is not great huh? I went down a similar path as you with [a201039](https://github.com/ustaxes/UsTaxes/pull/1098/commits/a2010394207ebdd854503bb03e90814ce40eb7fd). It looks like you struggled to get the types to work for migrations like I did, then figured out that it would require maintaining type definitions for each migration version, and then eventually figured it wasn't worth it. I had the same experience.

So I think it's fine to just revert to javascript style and assume we have no type information for the intermediate values.

ctSkennerton:
Yeah, started going down a rabbit hole and then got too busy to keep pushing on it. Thanks for merging in the latest changes. I'll have a look at why the tests are failing.

zakpatterson:
Working through this PR caught a ton of bugs. 

This PR includes #1152, merge #1152 first.

thegrims (changes requested):
image 
I think the datepicker should be initialized with a blank date for birth date rather than the current date

thegrims (changes requested):
Looks like taxpayers who are blind are eligible for an additional standard deduction, I see the standard deduction applied in the generated 1040, but I think the additional standard deduction for blindness is missing

> Additional Standard Deduction – You're allowed an additional deduction if you're age 65 or older at the end of the tax year. You're considered to be 65 on the day before your 65th birthday. You're allowed an additional deduction for blindness if you're blind on the last day of the tax year. For example, a single taxpayer who is age 65 and blind would be entitled to a basic standard deduction and an additional standard deduction equal to the sum of the additional amounts for both age and blindness.

https://www.irs.gov/taxtopics/tc551

ctSkennerton:
@thegrims thanks for the review. I've added in the additional logic for the additions to the standard deduction and fixed up the default date of birth on the tax payer form.

thegrims:
> @thegrims thanks for the review. I've added in the additional logic for the additions to the standard deduction and fixed up the default date of birth on the tax payer form.

Awesome, thanks for doing that!

thegrims (approved):

thegrims:
Thanks for your contribution!

PR #1104, opened 2022-04-12
Fix line 19 of 2021 1040

### LINKED ISSUE #1093: Schedule 8812

zakpatterson:
Remove ChildTaxWorksheet for 2021 (eliminated)

Move qualifying dependent logic to new (pseudo)worksheet.

Fixes #1093 

- Bugfix


zakpatterson:
This behavior is happening on ustaxes.org right now -- i'll work on a fix.

zakpatterson:
Ok, fixed that on master, should be fixed here now.

thegrims (approved):

PR #1119, opened 2022-04-15
More typescript eslint rules

zakpatterson:
* First commit: Adds the [typescript-eslint rules that are recommended, which consider type information](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/eslint-plugin#recommended-configs). This fixes numerous different potential bugs and points of confusion. For example:
    * Floating promises are explicitly marked. These are promises that are created where it's not possible to await them. This would always be a potential bug, except in the case of user events. `onClick` expects a method returning void, but often we want to do something asynchronously. So these cases are marked with a new function `intentionallyFloat`.
    * Many cases in tests where functions were declared as `async` but didn't `await` anything.
    * Created a new type for a `SoldAsset`, which requires a `closeDate` and a `closePrice`. Eliminates lots of unnecessary undefined checking.
    * Improves generic form elements so there's better typechecking around the values and names, based on the type of the form data coming in from context. 
    
  * Second commit: Adds a (not officially recommended but perfect for this project) [`no-unnecessary-condition` rule](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unnecessary-boolean-literal-compare.md).  Benefits:
    * We have a lot of fields that return `number` and lots that return `number | undefined`. The basic rules catch cases where we're treating a field incorrectly as a field returning number when it can be undefined. This rule makes sure we're not needlessly checking for undefined when the types say undefined is not possible. In the future if we remove `| undefined` from a field, the linter will find all the places where we were doing something like `?? 0`.
    * This also catches cases where we wrote something like: `x() ?? 0 > 0`! `??` is lower precedence than `>` so that expression for `x: () => A` would have type `A | boolean` and might not behave as expected. 

- Bugfix
- Build-related changes


zakpatterson:
Intermittent issue came up in build, present on master, added failing test as #1129.

This one should be ready.

thegrims (approved):

PR #1135, opened 2022-04-20
Add a new type ValidatedInformation, fix summary page lag

zakpatterson:
* Remove all situations in forms where we have to double check that `filingStatus` is defined, or that `primaryPerson` information is defined.

  * Precheck errors before rendering the Summary page. This fixes a noticeable rendering lag for the final page, as we were generating the entire 1040 to calculate any errors if they exist. Anything that can be an error should be checked before we attempt to buld the 1040.

- Bugfix
- Refactor


thegrims (approved):
LGTM

PR #1137, opened 2022-04-21
Fix Fica test is skipping 100% of the time

zakpatterson:
Marking a filtered test with a pre-condition shows that this test is never actually being run.

We probably need to [change the arbitraries](https://github.com/ustaxes/UsTaxes/blob/master/src/core/tests/arbitraries.ts#L104) to generate values in the domain this property is supposed to be testing

@ctSkennerton can you take a look? 

```
 FAIL  src/forms/Y2021/tests/fica.test.ts (133.231 s)
  fica
    ✕ should give SS refund based on filing status (126134 ms)

  ● fica › should give SS refund based on filing status

    Failed to run property, too many pre-condition failures encountered
    { seed: -476405088 }

    Ran 0 time(s)
    Skipped 10001 time(s)

    Hint (1): Try to reduce the number of rejected values by combining map, flatMap and built-in arbitraries
    Hint (2): Increase failure tolerance by setting maxSkipsPerRun to an higher value
    Hint (3): Enable verbose mode at level VeryVerbose in order to check all generated values and their associated status

```

- Bugfix

### DIFF
--- a/src/core/tests/arbitraries.ts
+++ b/src/core/tests/arbitraries.ts
@@ -60,6 +60,7 @@ const investmentResult = posNegCurrency(100000)
 const expense: Arbitrary<number> = posCurrency(10000)
 const interest: Arbitrary<number> = posCurrency(10000)
 const payment: Arbitrary<number> = fc.nat({ max: 100000 })
+const ssWitholding: Arbitrary<number> = fc.nat({ max: 10000 })
 
 const payerName: Arbitrary<string> = maxWords(3)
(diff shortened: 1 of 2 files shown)


ctSkennerton:
@zakpatterson I can't push onto this directly but I have a pull request against your fork. https://github.com/zakpatterson/UsTaxes/pull/93

zakpatterson:
Looks good, thanks @ctSkennerton !

PR #1152, opened 2022-04-27
Fix bug caused by reset of form with {}

zakpatterson:
When using an open/closable form, such as for dependents, W2s, 1099s, etc, after the form is submitted the first time, or the form is opened and then closed, on reopening the form will fail if there is a checkbox on the form and the checkbox is not selected on a subsequent use.

This is caused by the form being initialized with an object of default values including `{checkboxValue: false}`, and then reset with `{}`, so `checkboxValue` is `undefined` and submitted in the action undefined while it is required. Our AJV validation catches this and prevents the object with missing properties from being created in the redux store.

[Note this doc suggests that if reset is called then all the properties should be supplied](https://react-hook-form.com/api/useform/reset/), or the html form reset api is used. HTML reset api is not useful for us because the form container may be used without a `<form>` tag. Unfortunately `react-hook-form` doesn't make the `defaultValues` the form was created with accessible in context, so we have to create a new property to pass to `FormListContainer`.
 

- Bugfix

### DIFF
--- a/src/components/FormContainer.tsx
+++ b/src/components/FormContainer.tsx
@@ -14,7 +14,7 @@ import {
   Theme
 } from '@material-ui/core'
 import { Delete, Edit } from '@material-ui/icons'
-import { SubmitHandler, useFormContext } from 'react-hook-form'
+import { DefaultValues, SubmitHandler, useFormContext } from 'react-hook-form'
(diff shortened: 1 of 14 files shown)


zakpatterson:
I reviewed this, and it's required for #1098.

PR #1161, opened 2022-04-29
Add Support page and data anonymizer

zakpatterson:
Adds a new page for the user to submit feedback or bug reports. 

Allows the user to copy paste anonymized data, where all names, addresses, SSN, EIN, account numbers, etc are removed.

image

- Feature


thegrims (approved):
LGTM! This will be really helpful

PR #1718, opened 2023-04-07
Y2022: 8812: fix the 12yes/no error issue

### LINKED ISSUE #1717: The calculation of the credit for children and dependents in 1040 is wrong
Form 8812 line 12 yes/no logic error, which leads to the credit is not taken in while it actually should be.

keyonjie:
Fixes #1717 
The line 12yes/no logic is opposite, which will eventually lead to the credit error, here fix it.


thegrims (approved):
Thanks for the fix!

PR #2045, opened 2024-03-09
Fix investments after January 31, 2022 not being able to be entered

saiarcot895:
The investments entry form had a limitation that the max date was `new Date(2021, 12, 31)`. Because Javascript months are 0-indexed, this actually mapped to January 31, 2022. This also meant that investments bought or sold after this date couldn't be entered.

Fix this by having the max date be today. I think there's anyways logic to only use investments in the specified tax year for tax calculations. Also make sure that the date sold field has a date that isn't older than the date acquired field.

- Bugfix


thegrims:
Good catch!

thegrims (approved):

Issue #2059, opened 2024-03-18
Deleting your data 


robinf-scott:
do not want to utilize  this site. want my information deleted.

swelborn:
https://github.com/ustaxes/UsTaxes?tab=readme-ov-file#user-data

chenyuan99:
you can clear your local storage in 
<img width="542" alt="image" src="https://github.com/user-attachments/assets/935204d4-a08f-432f-beab-3b62edebd075">
But I think it would also be a good idea for us to create a new feature to cclear out all local storage in one click @

PR #2221, opened 2024-09-20
Feat add local storage clear button to give user more control and transparency

chenyuan99:
What kind of change does this PR introduce? (delete not applicable)

Feature
Docs
Fixes https://github.com/ustaxes/UsTaxes/issues/2059


thegrims:
Looks like this clears local storage but information is still persisted in the forms until the browser is refreshed, which might be confusing to some users. Could we add text asking them to refresh the page once cleared or similar?

chenyuan99:
It make sense, also we can use window.load to help user refresh directly

thegrims (approved):
Nice work! LGTM

PR #2318, opened 2025-01-05
doc: add badge for openssf best practice

chenyuan99:

- Docs


thegrims:
Can you add some information here about the significance of this badge?

chenyuan99:
I think it would be good for us to get a bit more clear way to see what license obligations from dependencies, maybe some depency licenses we might also need to include

Issue #2317, opened 2025-01-05
Checksums available?


kohded:
Do you provide checksums with the installers release? Can't seem to find it, but would be nice for the added security.

PR #2476, opened 2025-11-02
#2317 add checksum for release process

chenyuan99:
- Feature


chenyuan99:
I think since we are doing build release, it is  good practice to calculate sha256 to make sure the integrity of the build artficat is garanteed and in case some one in bad faith might to do supply-chain/dependency attack.

thegrims (approved):

PR #2565, opened 2026-06-14
fix: make combinePdfs linear to preserve page order and content

thegrims:
### Summary

This PR replaces the previous quadratic merge implementation in src/core/pdfFiller/pdfHandler.ts with a direct page-copy approach. The old code saved and reloaded PDFs repeatedly while combining them, which caused unnecessary work and risked order/content regressions.

### What changed

- `combinePdfs()` now creates a new `PDFDocument` and copies pages from each source PDF in sequence
- Preserves document order by iterating input PDFs in order and appending pages in page index order
- Eliminates repeated `save()` + `load()` cycles, reducing complexity from O(n^2) to O(n) for multi-document merges

### Test plan

- Installed dependencies with `npm install`
- Verified `src/tests/components/CreatePdf.test.tsx` passes
- Re-ran the same test file 10 times successfully with no failures

### Notes

This fix is focused on the merge helper; no UI changes were made.

### DIFF
--- a/src/core/pdfFiller/pdfHandler.ts
+++ b/src/core/pdfFiller/pdfHandler.ts
@@ -22,18 +22,14 @@ export const downloadPDF: PDFDownloader = async (url) => {
 export const combinePdfs = async (
   pdfFiles: PDFDocument[]
 ): Promise<PDFDocument> => {
-  const [head, ...rest] = await Promise.all(
-    pdfFiles.map(async (pdf) => PDFDocument.load(await pdf.save()))
-  )
+  const mergedPdf = await PDFDocument.create()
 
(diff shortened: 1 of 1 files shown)


src/core/pdfFiller/pdfHandler.ts
zakpatterson, line 1: Well that's a lot simpler

zakpatterson (approved):

src/core/pdfFiller/pdfHandler.ts
thegrims: Right? I was looking into our unit test timeouts and found this suggested in the pdf-lib docs

PR #2575, opened 2026-07-01
fix: prevent parseFormNumber from leaking NaN into tax calculations

Montana:
Hello US Taxes, 

I've came across a `NaN` issue, but was able to resolve it: 

`parseFormNumber` in `src/core/util.ts` is typed `number | undefined`, but it
returns `NaN` for non-numeric input:

```ts
export const parseFormNumber = (x: string | undefined): number | undefined => {
  if (x !== undefined && x.length > 0) {
    try {
      return parseFloat(x)
    } catch (e) {
      return undefined
    }
  }
  return undefined
}
```

`parseFloat` never throws — it returns `NaN` — so the `try/catch` is dead code, and inputs like `"abc"` or `"  "` return `NaN` instead of `undefined`.

This has real downstream effects:

- **`parseFormNumberOrThrow` stops throwing.** It guards with
  `res === undefined`, and since `NaN !== undefined`, it silently returns `NaN`  instead of throwing. It's used for W-2 fields (`income`, `fedWithholding`,
  `ssWages`, etc.) in `W2JobInfo.tsx`, so a bad value flows straight into the data model.
- **`NaN` poisons the return.** Once a `NaN` enters tax math it propagates through every calculation that touches it.
- **`?? 0` fallbacks don't catch it.** The box12 display does
  `parseFormNumber(x) ?? 0`; `??` only catches `null`/`undefined`, so bad input renders `NaN` rather than `0`.

## Fix

Drop the dead `try/catch` and normalize `NaN` to `undefined`:

```ts
export const parseFormNumber = (x: string | undefined): number | undefined => {
  if (x !== undefined && x.length > 0) {
    const parsed = parseFloat(x)
    return isNaN(parsed) ? undefined : parsed
  }
  return undefined
}
```

Valid inputs are unchanged (`"123"`, `"12.5"`, `"-4"`, `"1e3"`, `"0"`). `parseFormNumberOrThrow` now throws on non-numeric input as intended, and `?? 0` fallbacks resolve to `0`.

## Tests

Adds `src/core/tests/util.test.ts` covering valid parsing, empty/undefined input, and the non-numeric cases. The non-numeric and `orThrow` assertions fail on `master` and pass with this change.

Cheers,
Michael

### DIFF
--- a/src/core/tests/util.test.ts
+++ b/src/core/tests/util.test.ts
@@ -1,0 +1,32 @@
+import { parseFormNumber, parseFormNumberOrThrow } from '../util'
+
+describe('parseFormNumber', () => {
+  it('parses valid numeric strings', () => {
+    expect(parseFormNumber('123')).toBe(123)
+    expect(parseFormNumber('12.5')).toBe(12.5)
+    expect(parseFormNumber('0')).toBe(0)
+    expect(parseFormNumber('-4')).toBe(-4)
+    expect(parseFormNumber('1e3')).toBe(1000)
+  })
+
+  it('returns undefined for empty or absent input', () => {
+    expect(parseFormNumber('')).toBeUndefined()
+    expect(parseFormNumber(undefined)).toBeUndefined()
+  })
+
+  it('returns undefined (not NaN) for non-numeric input', () => {
+    expect(parseFormNumber('abc')).toBeUndefined()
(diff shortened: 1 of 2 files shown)


thegrims (approved):
Good catch! Thanks for this

